PR #4, opened 2022-11-18
Initial set up

moiskillnadne:
https://mindlogger.atlassian.net/browse/ML-115
https://mindlogger.atlassian.net/browse/ML-116
https://mindlogger.atlassian.net/browse/ML-117
https://mindlogger.atlassian.net/browse/ML-119
https://mindlogger.atlassian.net/browse/ML-120
https://mindlogger.atlassian.net/browse/ML-121


package.json
```diff
@@ -2,21 +2,44 @@
   "name": "mindlogger-rebuild",
   "private": true,
   "version": "0.0.0",
-  "type": "module",
+  "type": "commonjs",
   "scripts": {
     "dev": "vite",
```
BamMironov, line 7: Please add "npx" for all commands (example "npx vite") that refer to an npm package that is located in devDependencies to ensure the build will not fail during CI/CD processes.

src/app/App.tsx
```diff
@@ -0,0 +1,42 @@
+import { useState } from "react"
+import reactLogo from "../assets/react.svg"
+import "./App.css"
+import i18nManager from "./i18n"
+import Layout from "./providers/layout"
+import ReactQuery from "./providers/react-query"
+import Redux from "./providers/state"
```
BamMironov, line 7: Please organize the import's strategy.
For example, import { } from "react" goes first. Then go imports from npm dependencies. Then goes a section with absolute imports. Then goes a section with relative imports.  The last ones are css related.
Each section should be divided by an empty line;

Example:
image

You can choose whatever strategy you like but stick to one in every file.

vite.config.ts
```diff
@@ -1,7 +1,8 @@
 import { defineConfig } from 'vite'
 import react from '@vitejs/plugin-react'
+import eslint from 'vite-plugin-eslint'
 
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [react()]
+  plugins: [react(), eslint()]
```
BamMironov, line 7: Please use conditional config to distinguish between Dev and Prod. May it is not a good idea to execute eslint plugin during a production build.
https://vitejs.dev/config/#conditional-config

src/app/providers/react-query.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const ReactQuery = (props: ReactQueryProps) => {
+  return (
+    <Suspense>
+      <QueryClientProvider client={queryClient}>{props.children}</QueryClientProvider>
+    </Suspense>
+  )
+}
+
+export default ReactQuery
```
BamMironov, line 24: Don't forget to create a index.ts file and reexport all providers in order to import them from one place

src/app/App.tsx
```diff
@@ -0,0 +1,42 @@
+import { useState } from "react"
+import reactLogo from "../assets/react.svg"
+import "./App.css"
+import i18nManager from "./i18n"
+import Layout from "./providers/layout"
```
BamMironov, line 5: Layout component is not really a provider. It is just a component so it should be stored in another place.

src/app/providers/react-query.tsx
```diff
@@ -0,0 +1,24 @@
...
+  },
+})
+
+interface ReactQueryProps {
+  children: React.ReactNode
+}
+
+const ReactQuery = (props: ReactQueryProps) => {
+  return (
+    <Suspense>
```
BamMironov, line 18: Not sure that Suspense should be placed here. It should be imported and used in a file that bundles up all providers.

BamMironov (changes requested):

package.json
moiskillnadne: Done

src/app/App.tsx
moiskillnadne: Done

vite.config.ts
moiskillnadne: Done

src/app/providers/react-query.tsx
moiskillnadne: Done

src/app/App.tsx
moiskillnadne: Done

src/app/providers/react-query.tsx
moiskillnadne: Done

BamMironov (approved):

PR #17, opened 2022-12-27
Feature: forgot password page

moiskillnadne:


src/features/ForgotPassword/model/schemas.ts
```diff
@@ -0,0 +1,5 @@
+import { UserSchema } from "~/entities"
+import { z } from "zod"
```
BamMironov, line 2: Importing order

src/features/ForgotPassword/ui/index.tsx
```diff
@@ -0,0 +1,50 @@
+import { Alert, Container } from "react-bootstrap"
+import { useNavigate } from "react-router-dom"
+
+import { useForgotPasswordMutation } from "~/entities"
+import { BasicButton, BasicFormProvider, Input, isObjectEmpty, ROUTES, useCustomForm } from "~/shared"
+
+import { useForgotPasswordTranslation } from "../lib/useForgotPasswordTranslation"
+import { ForgotPasswordSchema, TForgotPasswordForm } from "../model/schemas"
+
+export const ForgotPasswordForm = () => {
```
BamMironov, line 10: I think in the scope of a feature a form should be in a separate file, e.g. `ForgotPasswordForm.tsx`, not in the `index.tsx`. Because an `index` file implies the feature (user's action) itself. And this case a feature is a set of ui and models.

src/features/Signup/ui/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+    mutate: signup,
+    isError,
+    error,
+  } = useSignupMutation({
+    onSuccess,
+  })
+
+  const onSignupSubmit = (data: TSignupForm) => {
+    if (data.confirmPassword !== data.password) {
+      return setError("confirmPassword", { type: "pattern", message: t("passwordsUnmatched") || "" })
```
BamMironov, line 48: Please use zod's `refine` function in scheme instead 
https://github.com/colinhacks/zod#refine

BamMironov (changes requested):

src/features/ForgotPassword/model/schemas.ts
moiskillnadne: Done

src/features/ForgotPassword/ui/index.tsx
moiskillnadne: Done

src/features/Signup/ui/index.tsx
moiskillnadne: Done

BamMironov (approved):

PR #18, opened 2022-12-29
Feature: Change password page

moiskillnadne:


src/entities/user/lib/authorization.service.ts
```diff
@@ -37,6 +44,21 @@ export class AuthorizationService {
...
+    const params = {
+      token: data.temporaryToken,
+    }
+
+    return this.httpService.GET(`/user/password/temporary/${data.userId}`, { params })
+  }
+
+  public updatePassword(data: IUpdatePasswordPayload) {
+    const headers = { "Girder-Token": data.token }
+    const params = { old: data.oldPassword, new: data.newPassword }
```
BamMironov, line 58: Please make sure that IUpdatePasswordPayload accepts { old, new } so that you don't have to make transformations inside the method.

BamMironov (changes requested):

src/entities/user/lib/authorization.service.ts
moiskillnadne: Done

BamMironov (approved):

PR #25, opened 2023-01-25
Feature: Applet list api / Interceptors / Secure Storage

moiskillnadne:


src/entities/user/model/user.slice.ts
```diff
@@ -0,0 +1,29 @@
...
+  fullName: null,
+  id: null,
+}
+
+const userSlice = createSlice({
+  name: "user",
+  initialState,
+  reducers: {
+    save: (state, action: PayloadAction<UserStore>) => {
+      return { ...state, ...action.payload }
```
BamMironov, line 20: Redux Toolkit uses Immer internally
https://redux-toolkit.js.org/usage/immer-reducers

please make use of it when writing reducers.

src/features/ChangePassword/ui/ChangePasswordForm.tsx
```diff
@@ -61,6 +61,8 @@ export const ChangePasswordForm = ({ title, token, email, onSuccessExtended }: C
   const approveErrorMessage = useMemo(() => approveError?.evaluatedMessage, [approveError])
   const updateErrorMessage = useMemo(() => updateError?.evaluatedMessage, [updateError])
 
+  const approveSuccessMessage = useMemo(() => (isApproveSuccess ? t("success") : null), [isApproveSuccess, t])
```
BamMironov, line 64: I didn't notice that during previous reviews but anyway: You don't need `useMemo` here. If it is a dead simple computation then we don't wrap it in useMemo

src/features/ProtectedRoute/ui/InactivityTracker.tsx
```diff
@@ -2,36 +2,40 @@ import { PropsWithChildren, useCallback, useEffect, useRef } from "react"
...
   const navigate = useNavigate()
+  const { clearUser } = userModel.hooks.useUserState()
 
-  const { clearUserAndAuth, auth } = useAuth()
-  const { mutate: logout } = useLogoutMutation()
+  const { mutate: logout } = useLogoutMutation({
+    onSuccess() {
+      clearUser()
+      userModel.secureTokensStorage.clearTokens()
+      navigate(ROUTES.login.path)
```
BamMironov, line 25: All of these need to be performed even if the logout request fails

src/features/Signup/ui/SignupForm.tsx
```diff
@@ -21,10 +23,20 @@ export const SignupForm = () => {
...
 
-  const { mutate: signup, error, isSuccess } = useSignupMutation()
+  const {
+    mutate: signup,
+    error,
+    isSuccess,
+    isLoading,
+  } = useSignupMutation({
+    onSuccess() {
+      reset()
```
BamMironov, line 37: Why?

src/shared/api/services/axios.ts
```diff
@@ -7,4 +14,50 @@ const axiosService = axios.create({
...
+      }
+
+      try {
+        const { data } = await authorizationService.refreshToken({ refreshToken: tokens?.refreshToken })
+
+        secureTokensStorage.setTokens(data.result)
+
+        config.headers!.Authorization = `${data.result.tokenType} ${data.result.accessToken}`
+      } catch (e) {
+        Promise.reject(e)
```
BamMironov, line 53: What happens if refresh token is expired?

BamMironov (changes requested):

src/entities/user/model/user.slice.ts
moiskillnadne: Done

src/features/ChangePassword/ui/ChangePasswordForm.tsx
moiskillnadne: Done

src/features/Signup/ui/SignupForm.tsx
moiskillnadne: Done

src/shared/api/services/axios.ts
moiskillnadne: Done

BamMironov (approved):

PR #58, opened 2023-03-26
feature: CheckboxItem component

moiskillnadne:


src/entities/activity/lib/types/item.ts
```diff
@@ -22,3 +39,52 @@ export type ItemCardButtonsConfig = {
...
+  timer: number | null
+  addScores: boolean
+  setAlerts: boolean
+  addTooltip: boolean
+  setPalette: boolean
+  additionalResponseOption: {
+    textInputOption: boolean
+    textInputRequired: boolean
+  }
+}
```
BamMironov, line 78: I'd propose refactoring some of the properties into separate types:

```
type ButtonsConfig = {
  removeBackButton: boolean;
  skippableItem: boolean;
};

type TimerConfig = {
  timer: number | null;
};

type AdditionalResponseConfig = {
  additionalResponseOption: {
    textInputOption: boolean;
    textInputRequired: boolean;
  };
};
```
They are very common among the other item DTOs. So you can use them to compose different types without writing too much boilerplate in the future

src/entities/activity/lib/types/item.ts
```diff
@@ -22,3 +39,52 @@ export type ItemCardButtonsConfig = {
   isSkippable?: boolean
   isNextDisable?: boolean
 }
+
+export type BaseItem<Config, ResponseValues> = {
```
BamMironov, line 43: It is better to name it just `ActivityItem` then

BamMironov (changes requested):

src/entities/activity/lib/types/item.ts
moiskillnadne: Done

PR #93, opened 2023-05-14
[M2-1844] Feature/encryption browserify

moiskillnadne:


src/entities/activity/lib/hooks/useSaveEncryptedAnswers.ts
```diff
@@ -0,0 +1,60 @@
...
+  answers: Array<AnswerTypesPayload>
+}
+
+type EncryptAnswerReturn = {
+  activityItemId: string
+  answer: string
+}
+
+export const useEncrypteAnswers = () => {
+  const { generateAesKey, encryptDataByKey, generateUserPublicKey } = useEncryption()
```
BamMironov, line 16: Please create the "createEncryptionService" factory that is returned `useEncryption`. Then use it in encrypteAnswers function. 
`createEncryptionService` is a function that accepts all necessary key as params and returns object with the "encrypt" method. The encrypt method should only accept JSON string. Key should be created internally during `createEncryptionService` execution

BamMironov (changes requested):

BamMironov (approved):

PR #96, opened 2023-05-17
Feature/m2 1527 Conditional logic

moiskillnadne:


src/entities/activity/lib/conditionalLogicBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    if (!condition) {
+      return true
+    }
+
+    const itemMapByName: ItemMapByName = {}
+
+    array.forEach(item => {
+      itemMapByName[item.name] = item
+    })
```
BamMironov, line 27: Please use [Array's .reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) instead.

src/entities/activity/lib/conditionalLogicBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+  public process(items: ActivityEventProgressRecord[]): ActivityEventProgressRecord[] {
+    return items.filter(this.conditionalLogicFilter)
+  }
+
+  private conditionalLogicFilter(
+    item: ActivityEventProgressRecord,
+    index: number,
+    array: ActivityEventProgressRecord[],
+  ): boolean {
+    const condition = item.conditionalLogic
```
BamMironov, line 17: please come up with a different name. e.g conditionalLogic

BamMironov (changes requested):

PR #192, opened 2023-08-18
M2 3002 add mixpanel events

DimitrySakalouski:


src/shared/utils/mixpanel.ts
```diff
@@ -0,0 +1,32 @@
+import mixpanel, { Dict } from "mixpanel-browser"
+
+const isProduction = process.env.REACT_APP_ENV === "PRODUCTION"
```
moiskillnadne, line 3: I have concerns that these variables won`t work.

The Vite allows to use variables with the prefix VITE_{anything}
And to get access to environmental variables you need to use `import.meta.env.{name-of-variable}`

More information about how variables work:
https://vitejs.dev/guide/env-and-mode.html#env-variables-and-modes

moiskillnadne (changes requested):

src/shared/utils/mixpanel.ts
BamMironov: Yeah, that's my copy-paste didn't work as I expected

moiskillnadne (approved):

moiskillnadne (approved):

PR #358, opened 2024-01-30
[M2-4788] Feature: "Next" button behavior update

moiskillnadne:
Resolve: [M2-4788](https://mindlogger.atlassian.net/browse/M2-4788)

[M2-4788]: https://mindlogger.atlassian.net/browse/M2-4788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/widgets/ActivityDetails/ui/AssessmentPassingScreen.tsx
+++ b/src/widgets/ActivityDetails/ui/AssessmentPassingScreen.tsx
@@ -13,7 +13,7 @@ import { ActivityCardItem, ItemCardButton, useTextVariablesReplacer } from "~/en
 import { appletModel } from "~/entities/applet"
 import { ActivityDTO, AppletDetailsDTO, AppletEventsResponse, RespondentMetaDTO } from "~/shared/api"
 import { Theme } from "~/shared/constants"
-import { NotificationCenter, useNotification } from "~/shared/ui"
+import { MuiModal, NotificationCenter, useNotification } from "~/shared/ui"
 import { useAppSelector, useCustomTranslation, useFlowType, usePrevious } from "~/shared/utils"
 
 type Props = {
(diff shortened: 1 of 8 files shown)


src/widgets/ActivityDetails/ui/AssessmentPassingScreen.tsx
```diff
@@ -145,11 +149,12 @@ export const AssessmentPassingScreen = (props: Props) => {
     }
 
     if (!hasNextStep) {
-      return onSubmit()
+      return setIsModalOpen(true)
+      // return onSubmit()
```
sultanofcardio, line 153: Should we remove this? Seems unnecessary to leave it around

```suggestion
```

sultanofcardio (approved):
Sorry for the delay. This looks good to me 👍 I'm not familiar enough with the codebase yet to comment on much besides functionality but I left you one minor comment.

Separately I found a scenario that I think should be addressed if we don't already have a ticket for it. In the scenario where there are 4 items in the flow, and item 4 is shown conditionally based on a correct answer being given for item 3, I think the progress bar at the top of the activity flow gives it away that the question being answered isn't the final one.

https://github.com/ChildMindInstitute/mindlogger-web-refactor/assets/14842108/bdbf840d-1299-4c23-b55c-0547b39bfb1f

In the video above, I'm able to use the completion of the progress bar to guess the correct answer. I believe this is something a participant may do if they're familiar with the platform, and it may influence their answers. This ticket was meant to remove one such method (showing Done vs Next), so I think we should have another ticket to address the progress bar

src/widgets/ActivityDetails/ui/AssessmentPassingScreen.tsx
moiskillnadne: Oh, I forgot to remove it. Thank you!

PR #371, opened 2024-02-09
M2-5007: Restart/resume feature

vmkhitaryanscn:
https://mindlogger.atlassian.net/jira/software/c/projects/M2/issues/M2-5007
Added restart/resume buttons to ActivityCard, ui updates, added translations

### DIFF
--- a/src/shared/ui/MuiModal/index.tsx
+++ b/src/shared/ui/MuiModal/index.tsx
@@ -83,6 +83,7 @@ export const MuiModal = (props: Props) => {
       {showCloseIcon && (
         <CloseIcon
           onClick={onHide}
+          data-testid="customized-dialog-close-icon"
           sx={{
             color: Theme.colors.light.onSurfaceVariant,
             marginLeft: "auto",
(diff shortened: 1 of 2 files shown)


src/entities/applet/model/hooks/useStartEntity.ts
```diff
@@ -23,6 +23,15 @@ export const useStartEntity = () => {
     )
   }
 
+  function removeActivityProgress(activityId: string, eventId: string): void {
+    dispatch(
```
BamMironov, line 27: I don't think it is necessary to delete activity/flow before starting a new one. The implementations of `flowStarted` and `activityStarted` reducers already imply creating a new object that is going to replace the previous one:
image

src/entities/applet/model/hooks/useStartEntity.ts
```diff
@@ -44,20 +53,31 @@ export const useStartEntity = () => {
...
 
     const flow = flows.find(x => x.id === flowId)!
+
     const flowActivities: string[] = flow.activityIds
 
     const firstActivityId: string = flowActivities[0]
 
+    if (shouldRestart) {
+      removeFlowActivitiesProgress(flowActivities, eventId)
+    }
```
BamMironov, line 71: As said above, I don't think we need to do it

BamMironov (changes requested):

src/entities/applet/model/hooks/useStartEntity.ts
BamMironov: It turned out that removeActivityProgress removes the state of in-progress activities, like step, answers, etc

BamMironov (approved):

PR #431, opened 2024-04-05
M2-6212: Allow direct linking to activity

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-6212](https://mindlogger.atlassian.net/browse/M2-6212)

This change allows directly linking to an activity by wrapping the `< AppletDetailsPage />` component in an `<AuthorizationGuard />`. I also created a new reusable component called `<LoginWithRedirect />` that, in combination with the `<AuthorizationGuard />`, automatically takes the user to the login page if they try to access a page while unauthenticated, and then redirects them to wherever they originally tried to access.

~I contemplated making this `<LoginWithRedirect />` component the default fallback for the `<AuthorizationGuard />`, as that made sense to me but I opted not to do that at this time. Let me know if you think that's something worth doing.~

~After initial review, I ultimately decided to make the `<LoginWithRedirect />` component the default fallback for the `<AuthorizationGuard />`~

I went back and forth on this but eventually decided to leave the `<AuthorizationGuard />` component without a default fallback, and avoid increasing the scope of the PR.

I also removed the `isInvitationFlow` parameter from the `useOnLogin` hook, and chose instead to rely on the existence of the `backRedirectPath` parameter, since it seems that's what `isInvitationFlow` was meant to do anyway.

### 📸 Screenshots

Before this change, if you tried to directly access an activity by URL, you would be taken to the applets screen after authenticating

https://github.com/ChildMindInstitute/mindlogger-web-refactor/assets/14842108/63d74989-fcef-4770-a0f8-80e719760fa0

Now, you'll be redirected to the activity, provided the URL parameters you passed are valid

https://github.com/ChildMindInstitute/mindlogger-web-refactor/assets/14842108/63180c9d-18f6-4ba0-a937-17add8727305

### 🪤 Peer Testing

> [!WARNING]
> This won't work using the preview environment due to the way it currently handles client side routing. You must test this locally

First authenticate, and navigate to an activity. Then copy the URL from the browser, log out, paste the URL again and go. You should be asked to log in, after which you are taken back to the activity

### DIFF
--- a/src/entities/user/model/hooks/useOnLogin.ts
+++ b/src/entities/user/model/hooks/useOnLogin.ts
@@ -5,7 +5,6 @@ import { ROUTES } from '~/shared/constants';
 import { Mixpanel, secureTokensStorage, useCustomNavigation, useEncryption } from '~/shared/utils';
 
 type Params = {
-  isInvitationFlow?: boolean | undefined;
   backRedirectPath?: string | undefined;
 };
 
@@ -40,8 +39,8 @@ export const useOnLogin = (params: Params) => {
     setUser({ id: user.id, email: user.email, firstName: user.firstName, lastName: user.lastName });
     secureTokensStorage.setTokens(tokens);
 
-    if (params.isInvitationFlow) {
-      navigate(params.backRedirectPath as string);
+    if (params.backRedirectPath !== undefined) {
+      navigate(params.backRedirectPath);
     } else {
       Mixpanel.track('Login Successful');
       Mixpanel.login(user.id);
--- a/src/pages/ActivityDetails/index.tsx
+++ b/src/pages/ActivityDetails/index.tsx
@@ -3,6 +3,8 @@ import { useParams } from 'react-router-dom';
 
 import { useCustomTranslation } from '~/shared/utils';
 import { ActivityDetailsContext, ActivityDetailsWidget } from '~/widgets/ActivityDetails';
+import { AuthorizationGuard } from '~/widgets/AuthorizationGuard';
+import { LoginWithRedirect } from '~/widgets/LoginWithRedirect';
 
 function ActivityDetailsPage() {
   const { appletId, activityId, eventId } = useParams();
@@ -13,11 +15,13 @@ function ActivityDetailsPage() {
   }
 
   return (
-    <Box display="flex" flex={1}>
-      <ActivityDetailsContext.Provider value={{ appletId, activityId, eventId, isPublic: false }}>
-        <ActivityDetailsWidget />
-      </ActivityDetailsContext.Provider>
-    </Box>
+    <AuthorizationGuard fallback={<LoginWithRedirect />}>
+      <Box display="flex" flex={1}>
+        <ActivityDetailsContext.Provider value={{ appletId, activityId, eventId, isPublic: false }}>
+          <ActivityDetailsWidget />
+        </ActivityDetailsContext.Provider>
+      </Box>
+    </AuthorizationGuard>
   );
 }
 
--- a/src/pages/ActivityDetails/index.tsx
+++ b/src/pages/ActivityDetails/index.tsx
@@ -4,6 +4,7 @@ import { useParams } from 'react-router-dom';
 import { useCustomTranslation } from '~/shared/utils';
 import { ActivityDetailsContext, ActivityDetailsWidget } from '~/widgets/ActivityDetails';
 import { AuthorizationGuard } from '~/widgets/AuthorizationGuard';
+import { LoginWithRedirect } from '~/widgets/LoginWithRedirect';
 
 function ActivityDetailsPage() {
   const { appletId, activityId, eventId } = useParams();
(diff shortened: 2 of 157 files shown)


src/entities/user/model/hooks/useOnLogin.ts
```diff
@@ -5,7 +5,6 @@ import { ROUTES } from '~/shared/constants';
 import { Mixpanel, secureTokensStorage, useCustomNavigation, useEncryption } from '~/shared/utils';
 
 type Params = {
-  isInvitationFlow?: boolean | undefined;
```
farmerpaul, line 8: :+1:

src/pages/ActivityDetails/index.tsx
```diff
@@ -13,11 +15,13 @@ function ActivityDetailsPage() {
   }
 
   return (
-    <Box display="flex" flex={1}>
-      <ActivityDetailsContext.Provider value={{ appletId, activityId, eventId, isPublic: false }}>
-        <ActivityDetailsWidget />
-      </ActivityDetailsContext.Provider>
-    </Box>
+    <AuthorizationGuard fallback={<LoginWithRedirect />}>
```
farmerpaul, line 18: I really think this should be the default `fallback` for AuthorizationGuard, and we should get rid of `AuthorizationButtons`. The interim screens adds zero value to the UX; you can already easily opt whether to log in/sign up from the Login screen. Still might make sense for someone to override the `fallback` prop in theory, so I guess leave that prop available.

farmerpaul (commented):
Looks great @sultanofcardio, I like your changes. I made one suggestion and had a minor question.

Just a reminder to update the JIRA task description & AC to align with the new requirements before passing it over to QA (I'd probably just do it myself rather than wait for Suzanne, but your call).

src/widgets/ActivityDetails/ui/ActivityDetails.tsx
```diff
@@ -27,36 +31,21 @@ export const ActivityDetailsWidget = () => {
...
-      <Box height="100vh" width="100%" display="flex" justifyContent="center" alignItems="center">
-        <span>
-          {context.isPublic ? t('additional.invalid_public_url') : error?.evaluatedMessage}
-        </span>
-      </Box>
-    );
-  }
+  if (!appletDetails || !activityDetails || !eventsRawData || isError) {
+    setTimeout(() => {
+      showErrorNotification(t('unabletoLoadActivity'));
```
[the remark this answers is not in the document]
sultanofcardio: That's a good point. I'm not so sure that error message is completely valid. It says *"This assessment is no longer accepting respondents"*, but it would just be an invalid parameter. I think a generic error message is better than an incorrect one, but I'll wait to hear from a SS team member about it

src/pages/ActivityDetails/index.tsx
sultanofcardio: Agreed. I'll make these changes unless I get pushback about it 👍

farmerpaul: Actually it occurred to me – it looks like the `AuthorizationButtons` component is passing the `redirectState` to the Signup screen, but I'm not sure if the same thing is happening when we bypass `AuthorizationButtons` and try to navigate to Signup from the Login screen. Can you confirm we're not losing redirect behaviour on Signup?

sultanofcardio: After double checking this, I realise that the login and sign up pages don't pass state to each other. I'm going to revert this change rather than try to resolve this, since it was outside the scope of this PR anyway

farmerpaul (approved):

moiskillnadne (changes requested):
All current changes look good. But when I tested it locally I found one bug. The root of the issue is when we go directly to activity we skip the step where we set the activity group in the store. The activity group data contains info about flow, timestamps and so on. When we want to submit our answers we use some data from the activity group state. So we should fix it somehow because without this data we can`t send answers.

The issue
Screenshot 2024-04-10 at 23 52 50

The slice in the store
Screenshot 2024-04-10 at 23 53 27

The place in the code where we use this data. [/src/widgets/ActivityDetails/model/hooks/useAnswers.ts]
Screenshot 2024-04-10 at 23 53 51

sultanofcardio:
@moiskillnadne thanks for the review. I believe I've fixed the error you identified. Please review again

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
```diff
@@ -100,6 +101,7 @@ export const ActivityCard = ({ activityListItem }: Props) => {
       {
         onSuccess(data) {
           const activity = data.data.result;
+          console.log('Data fetched:', activity);
```
moiskillnadne, line 104: Please remove console log here

moiskillnadne (approved):
Good. Now it`s working. I left one comment about the console log but generally approve. Thank you!

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
sultanofcardio: Done

PR #437, opened 2024-04-16
M2-5929: Add LaunchDarkly SDK

ChaconC:
- [x] Related documentation has been added / updated

### 📝 Description

Adds LaunchDarkly SDK to web app.
Initial implementation with no targeting rules until these get defined in a later task.
For this reason, all flags will be enabled based on the environment, e.g. `DEV` has all flags turned on and `TEST` or `PROD` have all flags disabled.

🔗 [Jira Ticket M2-5929](https://mindlogger.atlassian.net/browse/M2-5929)


README.md
```diff
@@ -98,6 +98,7 @@ See the documentation on [Vitest](https://vitest.dev/guide/) for more informatio
 | `VITE_SECURE_LOCAL_STORAGE_PREFIX` | yes | `ML_SECURE` | Secure local storage prefix |
 | `VITE_IV_LENGTH` | yes | `16` | Encryption initialization vector length |
 | `VITE_BUILD_VERSION` | yes | `dev-build` | Build version |
+| `VITE_LAUNCHDARKLY_CLIENT_ID` | yes | null | LaunchDarkly environment ID |
```
sultanofcardio, line 101: Does it make sense to just set this to the value in the `.env.example` file?

https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/862f1125bfc94e73fa8652b2a8faaff50b4c447b/.env.example#L9

sultanofcardio (approved):
Looks good

farmerpaul (approved):
Looks like it's working fine! Code looks great, no issues there. I got it working only after remembering to add `VITE_LAUNCHDARKLY_CLIENT_ID` to `.env`. 😛 (Maybe good to include the Peer Testing section for more forgetful people… like me. 😉)

README.md
farmerpaul: Hmm… good point. In the admin app, the README is not matching `.env.example`. But I'm wondering if maybe we shouldn't be exposing our LaunchDarkly API key to the public at all? i.e. should it not live in `.env.example`?

farmerpaul: Sure… I guess my question is more, does it make sense for members of the public, if running the software in their own server environments, to be using CMI's LD account? Should their traffic count towards CMI's LD resource usage? I suppose access (by IP/domain) could be limited on the LD side if that ever became an issue.

sultanofcardio: Oh. I don't think that's a thing that's supported. MindLogger does have an arbitrary server setup, but it's just for the backend. If someone where to deploy this FE app somewhere, then I guess that could become a concern. Maybe something to bring up in Slack 🤔

ChaconC: I don't see any such config over on LD's side.

I guess it does make sense to add this key since it's already in `.env.example` 👍

PR #438, opened 2024-04-18
M2-6042: Validate Params from Admin to Web App

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-6042](https://mindlogger.atlassian.net/browse/M2-6042)

This PR builds on #431 and introduces validation of the following query parameters:

- `startActivityOrFlow` - An activity or Flow ID
- `respondentId` - The respondent taking the assessment
- `subjectId` - The subject who the assessment is about

If one of these parameter is omitted, then no validation takes place and the app behaves as it does today with one exception: the activity or flow is not started. Starting the activity/flow automatically now requires the provision of all three parameters.

If they are all provided, then the following validation is performed:
- `startActivityOrFlow` identifies a valid activity
- `respondentId` matches the currently authenticated user
- `subjectId` corresponds to a real subject
- The currently authenticated user has one of these roles in the applet
  - Super Admin
  - Owner
  - Manager

If all the validation passes, then the activity/flow is started automatically and the context is stored in a multi-informant redux store.

### 📸 Screenshots

There isn't really anything to show, besides error banners I guess 🤷

### 🪤 Peer Testing

> [!NOTE]
> You'll need to test this change locally, as it depends on backend changes that haven't yet been deployed. Use the [feature/multiinformant-metapod](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/tree/feature/multiinformant-metapod) branch for your API

> [!NOTE]
> This feature depends on the `enableMultiInformant` feature flag. If it it not already enabled, you can manually enable it [here](https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/9ec96b45196bf8f1eb377850d8d225e51323b559/src/shared/utils/hooks/useLaunchDarkly.ts#L22-L30) like so:
>
> `features.enableMultiInformant = true;`

There are quite a few scenarios that need to be tested, so buckle up. For this test, you'll need the following set up in your workspace:
- An owner
- A manager
- An editor
- A reviewer
- A limited account participant

You will also need at least one applet with at least one activity.

Make note of the **user ID** and **subject ID** for each of the users above, as well as the **applet ID** and **activity ID** then using the following URL template:

`https://localhost:5173/protected/applets/{appletId}/?startActivityOrFlow={activityId}&respondentId={sourceUserId}&subjectId={targetSubjectId}`

Substitute for each of following scenarios, and validate the outcome:

Scenario | Source (user ID) | Target (subject ID) | Outcome
-- | -- | -- | --
[...]

PR #441, opened 2024-04-22
M2-6044: Submit Take Now Params

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-6044](https://mindlogger.atlassian.net/browse/M2-6044)

This PR updates the code that submits the activity responses to `/answers` to include the values for `target_subject_id` and `source_subject_id`. The values are only included if the `enableMultiInformant` feature flag is enabled.

### 📸 Screenshots

N/A

### 🪤 Peer Testing

> [!NOTE]
> This feature depends on the `enableMultiInformant` feature flag. If it it not already enabled, you can manually enable it [here](https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/9ec96b45196bf8f1eb377850d8d225e51323b559/src/shared/utils/hooks/useLaunchDarkly.ts#L22-L30) like so:
>
> `features.enableMultiInformant = true;`

> [!NOTE]
> There's an in-progress API patch that this PR depends on. If https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1264 is still open when you are reviewing this PR, please use that branch as the basis for your API. Otherwise, you should use the `feature/multiinformant-metapod` branch

Follow the peer testing steps in #438 and complete the activity that is opened with the network tab of dev tools open.
Look for the answer submission request and confirm that it includes the two fields `sourceSubjectId` and `targetSubjectId`:

Screenshot 2024-04-22 at 3 44 33 PM

### ✏️ Notes

N/A

### DIFF
--- a/src/features/TakeNow/ui/ValidateTakeNowParams.tsx
+++ b/src/features/TakeNow/ui/ValidateTakeNowParams.tsx
@@ -14,7 +14,8 @@ function ValidateTakeNowParams({
   respondentId,
 }: TakeNowParams) {
   const { showErrorNotification } = useNotification();
-  const { initiateTakeNow, isInMultiInformantFlow } = appletModel.hooks.useMultiInformantState();
+  const { initiateTakeNow, isInMultiInformantFlow, getMultiInformantState } =
+    appletModel.hooks.useMultiInformantState();
 
   const { isError, isLoading, isSuccess, error, data } = useTakeNowValidation({
     appletId,
@@ -27,7 +28,12 @@ function ValidateTakeNowParams({
     if (isSuccess && data) {
       const { sourceSubjectId, targetSubjectId } = data;
 
-      if (!isInMultiInformantFlow()) {
+      const multiInformantState = getMultiInformantState();
+      if (
+        !isInMultiInformantFlow() ||
+        sourceSubjectId !== multiInformantState.sourceSubjectId ||
+        targetSubjectId !== multiInformantState.targetSubjectId
+      ) {
         initiateTakeNow({ sourceSubjectId, targetSubjectId });
       }
     }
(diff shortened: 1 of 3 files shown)


ChaconC:
Which branch of the backend do I need to test this with? I'm using `feature/multiinformant-metapod` and it's rejecting the request:
CleanShot 2024-04-26 at 12 02 27
CleanShot 2024-04-26 at 12 03 02

The params are added as expected
CleanShot 2024-04-26 at 12 03 38

sultanofcardio:
> Which branch of the backend do I need to test this with? I'm using `feature/multiinformant-metapod` and it's rejecting the request:

@ChaconC what's in the body of the error response?

ChaconC:
> > Which branch of the backend do I need to test this with? I'm using `feature/multiinformant-metapod` and it's rejecting the request:
> 
> @ChaconC what's in the body of the error response?

`"Subject 7d00095c-cc5b-442b-bcbe-a656d57f50db not found"`

Looks like a problem when using take now on a Limited account. Full accounts work ok

sultanofcardio:
> Looks like a problem when using take now on a Limited account.
> Full accounts work ok

I'm not sure what the behaviour of the API should be in that case, but this flow doesn't support using a limited account as the `sourceSubjectId`. So that's a scenario we can ignore for now

ChaconC (approved):

farmerpaul (commented):
Hey @sultanofcardio, I'm currently struggling to get this to work. There may be some user error on my part, so I'm wading through that currently. But I've also discovered that it seems `sourceSubjectId` may not always be getting correctly initialized on the `multiInformantState` when the quiz starts (so not related to the code in this PR). It's time-consuming to figure out what's going on, so please bear with me. I just wanted to let you know before you merge so that in case I've discovered a bug, you can include the fix in this PR.

I'll keep you posted!

farmerpaul (approved):
After troubleshooting and figuring creating a workaround for a BE bug as mentioned in Slack, and testing again with your changes, all tests are now working. Thanks @sultanofcardio! 💪🏻

ChaconC (approved):

PR #442, opened 2024-04-23
M2-6109: [FE] [Web App] Header - Update to Include Subject and Respondent

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-6109](https://mindlogger.atlassian.net/browse/M2-6109)

This PR creates a `MultiInformantBadge` component that is shown during the completion of an activity in a multi-informant context. The component displays itself when the following are true:
- The `enableMultiInformant` feature flag is turned on
- The multi-informant state has been set. This state is set during the validation of multi-informant parameters in the `ValidateTakeNowParams` component.

#### Header badge with respondent hovered

multi-informant-badge-respondent

#### Header badge with subject hovered

multi-informant-badge-subject

### 🪤 Peer Testing

> [!NOTE]
> This feature depends on the `enableMultiInformant` feature flag. If it it not already enabled, you can manually enable it [here](https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/9ec96b45196bf8f1eb377850d8d225e51323b559/src/shared/utils/hooks/useLaunchDarkly.ts#L22-L30) like so:
>
> `features.enableMultiInformant = true;`

For this test, you'll need the following set up in your workspace:

An owner
A limited account participant
You will also need at least one applet with at least one activity.

#### Multi-Informant case

Make note of the owner **user ID** and the limited account participant **subject ID**, as well as the applet ID and activity ID. Then format them according to this URL template:

https://localhost:5173/protected/applets/{appletId}/?startActivityOrFlow={activityId}&respondentId={ownerUserId}&subjectId={limitedAccountSubjectId}

Launch that URL (and authenticate if required), then observe the activity layout. You should see the badge displayed as expected.

#### Regular flow

From the list of activities in the applet, launch one of the activities. You should observe that the badge is not displayed.

### ✏️ Notes

N/A

### DIFF
--- a/src/features/TakeNow/ui/MultiInformantBadge.tsx
+++ b/src/features/TakeNow/ui/MultiInformantBadge.tsx
@@ -1,0 +1,36 @@
+import { useMultiInformantState } from '~/entities/applet/model/hooks';
+import { MultiInformantBadgeTile } from '~/features/TakeNow/ui/MultiInformantBadgeTile';
+import { Theme } from '~/shared/constants';
+import { Box } from '~/shared/ui';
+
+export const MultiInformantBadge = () => {
+  const { getMultiInformantState } = useMultiInformantState();
+
+  const multiInformantState = getMultiInformantState();
+
+  if (
+    multiInformantState.sourceSubject === undefined ||
+    multiInformantState.targetSubject === undefined ||
+    multiInformantState.sourceSubject.id === multiInformantState.targetSubject.id
+  ) {
+    return null;
+  }
+
+  const { sourceSubject, targetSubject } = multiInformantState;
+
+  return (
+    <Box
+      display="flex"
+      padding="4px"
+      alignItems="center"
+      gap="4px"
+      width="192px"
+      height="40px"
+      borderRadius="8px"
+      border={`1px solid ${Theme.colors.light.surfaceVariant}`}
+    >
+      <MultiInformantBadgeTile type="Respondent" subject={sourceSubject} />
+      <MultiInformantBadgeTile type="Subject" subject={targetSubject} />
+    </Box>
+  );
+};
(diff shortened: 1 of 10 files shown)


src/features/TakeNow/ui/MultiInformantBadge.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const { sourceSubject, targetSubject } = getMultiInformantState();
+
+  if (!sourceSubject || !targetSubject) {
+    return null;
+  }
+
+  return (
+    <Box
+      display="flex"
+      padding="4px"
```
farmerpaul, line 24: 🦅 Although Figma says 4px padding, it's also got an "inner-aligned" border, which is a Figma-specific trick that doesn't actually map directly to CSS. In Figma, the 1px border is taking up 1px of the 4px padding, leaving 3px padding left:
```suggestion
      padding="3px"
```

farmerpaul (approved):
Looks great @sultanofcardio! Works perfectly. Just left some nitpicky 🦅 comments.

src/features/TakeNow/ui/MultiInformantBadge.tsx
sultanofcardio: Incredible. Thanks for the tip, I'll keep an eye out for this in the future

src/features/TakeNow/ui/MultiInformantBadge.tsx
```diff
@@ -0,0 +1,36 @@
...
+      alignItems="center"
+      gap="4px"
+      width="192px"
+      height="40px"
```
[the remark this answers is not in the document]
sultanofcardio: I actually just realised what you meant. The calculated size ends up being smaller than 40px when I specify this. Good catch

PR #456, opened 2024-05-08
hotfix(state-access): Guards was added

moiskillnadne:
### 📝 Description

🔗 [Jira Ticket M2-6534](https://mindlogger.atlassian.net/browse/M2-6534)

One of the new features includes state changes but at some specific moments, the application had the old version of the state. Guard was added to validate that the required state exists.

Changes include:

- Multiinformant feature

### DIFF
--- a/src/features/TakeNow/ui/ValidateTakeNowParams.tsx
+++ b/src/features/TakeNow/ui/ValidateTakeNowParams.tsx
@@ -31,13 +31,13 @@ function ValidateTakeNowParams({
       const multiInformantState = getMultiInformantState();
       if (
         !isInMultiInformantFlow() ||
-        sourceSubject.id !== multiInformantState.sourceSubject?.id ||
-        targetSubject.id !== multiInformantState.targetSubject?.id
+        sourceSubject.id !== multiInformantState?.sourceSubject?.id ||
+        targetSubject.id !== multiInformantState?.targetSubject?.id
       ) {
         initiateTakeNow({ sourceSubject, targetSubject });
       }
     }
-  }, [data, initiateTakeNow, isInMultiInformantFlow, isSuccess]);
+  }, [data, getMultiInformantState, initiateTakeNow, isInMultiInformantFlow, isSuccess]);
 
   useEffect(() => {
     if (error) {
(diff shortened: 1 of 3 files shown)


src/features/TakeNow/ui/ValidateTakeNowParams.tsx
```diff
@@ -31,13 +31,13 @@ function ValidateTakeNowParams({
...
-        sourceSubject.id !== multiInformantState.sourceSubject?.id ||
-        targetSubject.id !== multiInformantState.targetSubject?.id
+        sourceSubject.id !== multiInformantState?.sourceSubject?.id ||
+        targetSubject.id !== multiInformantState?.targetSubject?.id
       ) {
         initiateTakeNow({ sourceSubject, targetSubject });
       }
     }
-  }, [data, initiateTakeNow, isInMultiInformantFlow, isSuccess]);
+  }, [data, getMultiInformantState, initiateTakeNow, isInMultiInformantFlow, isSuccess]);
```
sultanofcardio, line 40: I believe we should leave the dependency arrays as-is. Since these changes are unrelated to the bug fix, they can be added to dev instead

src/widgets/ActivityDetails/model/hooks/useAnswers.ts
```diff
@@ -155,18 +155,20 @@ export const useAnswer = (props: Props) => {
...
+      props.activityId,
       props.eventId,
       props.eventsRawData,
-      props.flowId,
+      encryptPayload,
+      getGroupProgress,
+      featureFlags.enableMultiInformant,
+      generateUserPrivateKey,
       getMultiInformantState,
+      isInMultiInformantFlow,
```
sultanofcardio, line 171: Same here. Can we revert the changes to the dependency array?

src/features/TakeNow/ui/ValidateTakeNowParams.tsx
moiskillnadne: Thank you, makes sense. Done

src/widgets/ActivityDetails/model/hooks/useAnswers.ts
moiskillnadne: Done

sultanofcardio (approved):

PR #468, opened 2024-05-28
feat(mi): Consume API take now validation endpoint (M2-6746) (M2-6856)

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-6746](https://mindlogger.atlassian.net/browse/M2-6746)
🔗 [Jira Ticket M2-6856](https://mindlogger.atlassian.net/browse/M2-6856)

This PR adds the `useValidateMultiInformantAssessmentQuery` query, which is based on the `/answers/applet/{applet_id}/multiinformant-assessment/validate` API endpoint implemented in https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1316. This moves the majority of the take-now parameter validation to the API instead of being done solely on the frontend.

There is no functional difference between the two implementations, so you may refer to https://github.com/ChildMindInstitute/mindlogger-admin/pull/1728 for testing guidance

### 📸 Screenshots

N/A

### 🪤 Peer Testing

Please refer to https://github.com/ChildMindInstitute/mindlogger-admin/pull/1728 (specifically the table at the end of the description) for testing the existing functionality.

In addition, please use the URL template `https://localhost:5173/protected/applets/{appletId}?startActivityOrFlow={activityId}&sourceSubjectId={sourceSubjectId}&targetSubjectId={targetSubjectId}&respondentId={ownerUserId}` to test the following scenarios:
- Invalid applet ID
- Invalid activity ID
- Invalid source subject
- Invalid target subject
- User with insufficient permission (e.g. coordinator user)

### ✏️ Notes

The issue [M2-6856](https://mindlogger.atlassian.net/browse/M2-6856) was raised after this PR was up for review. This PR inadvertently fixes it, so I'm adding the issue number to the PR title and description.

[M2-6856]: https://mindlogger.atlassian.net/browse/M2-6856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/entities/answer/api/index.ts
+++ b/src/entities/answer/api/index.ts
@@ -1,0 +1,1 @@
+export * from './useValidateMultiInformantAssessmentQuery';
--- a/src/entities/answer/api/useValidateMultiInformantAssessmentQuery.ts
+++ b/src/entities/answer/api/useValidateMultiInformantAssessmentQuery.ts
@@ -1,0 +1,23 @@
+import { answerService, QueryOptions, ReturnAwaited, useBaseQuery } from '~/shared/api';
+
+type FetchFn = typeof answerService.validateMultiInformantAssessment;
+type Options<TData> = QueryOptions<FetchFn, TData>;
+type OptionalArgs<TData = ReturnAwaited<FetchFn>> = {
+  sourceSubjectId?: string;
+  targetSubjectId?: string;
+  activityOrFlowId?: string;
+  options?: Options<TData>;
+};
+
+export const useValidateMultiInformantAssessmentQuery = <TData = ReturnAwaited<FetchFn>>(
+  appletId: string,
+  optionalArgs?: OptionalArgs<TData>,
+) => {
+  const { options, ...queryParams } = optionalArgs || {};
+
+  return useBaseQuery(
+    ['multiInformantAssessmentDetails', { appletId }],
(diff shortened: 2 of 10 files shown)


farmerpaul (approved):
Code looks good, and tests all ran as expected, with a couple pretty minor things worth noting (and this might be preexisting behaviour, not strictly related to your changes):

- When I provide an invalid applet ID, I get this full-screen error message after a delay of several seconds: "This assessment is no longer accepting respondents"; an error message about invalid applet ID might be more useful, and shown as a banner instead, but if that's too much work then no need to bother.
- When I provide an invalid activity ID, I'm just taken to the list of activities with no error message; similarly, I'd expect some kind of banner error message.

sultanofcardio:
Thanks for the review @farmerpaul. I agree with your assessment regarding the invalid applet ID. That was pre-existing behaviour that I maintained because the fallback page is the "applet page" that shows the list of activities. I think modifying this to fall back to the home page in that case shouldn't be too difficult.

Also, the invalid activity ID should produce an error message, so that's a bug. Let me check that out

sultanofcardio:
@farmerpaul fixed it!

PR #469, opened 2024-05-31
feat(mi): Update mismatched respondent error message copy (M2-6667)

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-6667](https://mindlogger.atlassian.net/browse/M2-6667)

This PR adds an extra error message (`mismatchedRespondent`) to be shown in case the take now `respondentId` does not match the ID of the currently logged-in user. 

The error message it replaces (`invalidRespondent`) is currently shown in 4 scenarios:
- Respondent/logged-in user mismatch (the scenario this ticket addresses)
- Logged-in user is not an owner/manager
- Logged-in user does not have permission to provide answers to the applet
- Logged-in user does not have permission to fetch the details of the source/target subjects

That error message continues to be shown in those other scenarios

#### Before

image

#### After

image

### 🪤 Peer Testing

1. Initiate the take now flow in the respondent web app with an invalid `respondentId` value.
2. Observe the error message

### ✏️ Notes

N/A

### DIFF
--- a/src/i18n/en/translation.json
+++ b/src/i18n/en/translation.json
@@ -289,6 +289,7 @@
         "subjectOfResponses": "Subject of Responses"
       },
       "invalidRespondent": "You are not authorized to complete this assessment on behalf of this subject.",
+      "mismatchedRespondent": "There's a mismatch between the expected responder and the signed-in web app user. Log out of the web app and login with the responder’s login to initiate Take Now again.",
       "invalidSubject": "This subject does not exist.",
       "invalidActivity": "This activity does not exist.",
       "invalidApplet": "This applet does not exist.",
--- a/src/i18n/en/translation.json
+++ b/src/i18n/en/translation.json
@@ -289,7 +289,7 @@
         "subjectOfResponses": "Subject of Responses"
       },
       "invalidRespondent": "You are not authorized to complete this assessment on behalf of this subject.",
-      "mismatchedRespondent": "There's a mismatch between the expected responder and the signed-in web app user. Log out of the web app and login with the responder’s login to initiate Take Now again.",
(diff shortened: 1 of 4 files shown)


src/i18n/en/translation.json
```diff
@@ -289,6 +289,7 @@
         "subjectOfResponses": "Subject of Responses"
       },
       "invalidRespondent": "You are not authorized to complete this assessment on behalf of this subject.",
+      "mismatchedRespondent": "There's a mismatch between the expected responder and the signed-in web app user. Log out of the web app and login with the responder’s login to initiate Take Now again.",
```
farmerpaul, line 292: > and login with the responder’s login to initiate Take Now again.

My only issue with this copy is that it seems to imply that that logging out and logging into the responder's login re-initiates Take Now. But it doesn't cause Take Now to be re-initiated after you log in. You still have to manually close the tab and redo Take Now as an additional manual step. So I would probably reword this as:

`There's a mismatch between the expected responder and the signed-in web app user. Log out of the web app and log in with the responder’s login, then re-initiate Take Now.`

farmerpaul (approved):
Works fine, just a comment on the copy which I think is still misleading.

src/i18n/en/translation.json
sultanofcardio: @farmerpaul I agree with you. But this copy came from the ticket. Can you leave that feedback on the ticket please and tag Cari?

sultanofcardio: @farmerpaul I applied the copy change. Thanks for the suggestion 👌 I decided to reword the last bit from _"admin panel"_ to _"Admin App"_, since that's the wording used on the pop-up at the end of the activity

PR #471, opened 2024-06-03
(No-task) The condition was added to prevent access to undefined state

moiskillnadne:
### 📝 Description

The small bug was fixed without a ticket. The cause of the bug is that the application tried to access the undefined part of the loris integration state.

Changes include:

- useIntegrationSync

### DIFF
--- a/src/widgets/ActivityGroups/model/hooks/useIntegrationsSync.ts
+++ b/src/widgets/ActivityGroups/model/hooks/useIntegrationsSync.ts
@@ -14,8 +14,10 @@ export const useIntegrationsSync = ({ appletDetails }: Props) => {
   const consents = useAppSelector(appletModel.selectors.selectConsents);
 
   useEffect(() => {
-    const appletIntegrationService = new AppletIntegrationsService(consents, dispatch);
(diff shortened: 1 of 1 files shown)


sultanofcardio (changes requested):
Also account for it on this line

https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/69db2b12a5e2bd6d02273bf6f15dbd13dfd58a50/src/widgets/ActivityDetails/model/hooks/useAnswers.ts#L41

```typescript
const appletConsents = consents?.[props.applet.id] ?? null;
```

moiskillnadne:
> Also account for it on this line
> 
> https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/69db2b12a5e2bd6d02273bf6f15dbd13dfd58a50/src/widgets/ActivityDetails/model/hooks/useAnswers.ts#L41
> 
> ```ts
> const appletConsents = consents?.[props.applet.id] ?? null;
> ```

Thank you. Fixed

sultanofcardio (approved):

PR #477, opened 2024-06-10
feat: migrate from Notifications to Banners (M2-6946)

farmerpaul:
- [x] OSS packages added to MindLogger [open source credit page](https://mindlogger.atlassian.net/jira/servicedesk/projects/MLA/knowledge/articles/340623543?spaceKey=MLA)

### 📝 Description

🔗 [Jira Ticket M2-6946](https://mindlogger.atlassian.net/browse/M2-6946)

This PR replaces the `NotificationCenter`, `useNotification` hook, and related infrastructure with `Banners`, `useBanners`, adapted from the Admin App with only slight changes to align with the Web App's code structure.

Includes:
- Redux Toolkit slice to manage global state of banners model
- `Banner` base component (replaces `Notification` component)
  - Same as in Admin App, component is based on MUI Alert component, and styled in call to `createTheme`
  - Added `useWindowFocus` hook (copied verbatim from Admin App) required for enhanced a11y/UX
- `Banners` container component (replaces `NotificationCenter` component)
-  Four custom banners: `SuccessBanner`, `WarningBanner`, `ErrorBanner`, and `InfoBanner`
- `useBanners` hook that provides convenience functions for adding/removing banners (replaces `useNotification` hook)
- `useSessionBanners` hook, adapted from Admin App, which removes any displayed banners upon logout
- Tests for `Banner` and `Banners` components
  - Required adding `@testing-library/user-event` package and upgrading `@vitejs/plugin-react` to the latest version to fix [a bug](https://github.com/vitejs/vite-plugin-react/pull/79) with executing tests that involved rendering React components
> [!TIP]
> It may be easier to review this PR [commit-by-commit](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/477/commits) due to the number of changes.

### 📸 Screenshots

| Before                      | After                                 |
| Screenshot 2024-06-10 at 3 10 12 PM | Screenshot 2024-06-10 at 3 13 42 PM |
| Screenshot 2024-06-10 at 3 04 15 PM | Screenshot 2024-06-10 at 3 00 21 PM |

### 🪤 Peer Testing

**Requires `yarn install`**

1. Log in with a valid email but **invalid password**.
    **Expected outcome:** An error banner should render with updated styling, including a close button.
2. Wait 5 seconds.
    **Expected outcome:** Banner should disappear.
3. Repeat step 1, then press the **close button** before the banner disappears.
    **Expected outcome:** Banner should disappear.
4. Repeat step 1, then hover mouse cursor over the banner and wait more than 5 seconds.
    **Expected outcome:** Banner should remain visible.
5. Repeat step 1, then make the window/tab lose focus for more than 5 seconds. Refocus the window.
    **Expected outcome:** Banner should still be visible.
6. Wait 5 seconds with the cursor not hovering over the banner.
    **Expected outcome:** Banner should disappear.
7. Initiate Take Now on an activity, but log into the web app using an admin account that is different from the account selected for inputting the answers.
    **Expected outcome:** Error banner should should render with updated styling.
8. Before banner disappears, log out of the Web App.
    **Expected outcome:** Error banner should disappear.


src/pages/Login/index.tsx
```diff
@@ -26,10 +26,7 @@ function LoginPage() {
 
     if (location.state?.isPasswordReset) {
       // Shown for 5 seconds
-      showSuccessNotification(t('passwordResetSuccessful'), {
-        duration: 5000,
-        allowDuplicate: false,
-      });
+      addSuccessBanner({ children: t('passwordResetSuccessful'), duration: 5000 });
```
farmerpaul, line 29: `allowDuplicate` is no longer relevant with the new infrastructure as duplicates of the same banner key is impossible.

src/shared/utils/hooks/useSessionBanners/useSessionBanners.ts
```diff
@@ -0,0 +1,19 @@
...
+import { userModel } from '~/entities/user';
+
+export const useSessionBanners = () => {
+  const { isAuthorized } = userModel.hooks.useAuthorization();
+  const { removeAllBanners } = useBanners();
+
+  const prevIsAuthorized = useRef(isAuthorized);
+
+  useEffect(() => {
+    if (prevIsAuthorized.current !== isAuthorized && !isAuthorized) {
```
ChaconC, line 13: Wondering why the check for `!isAuthorized` feels like we should also clear notifications when login is successful.

CleanShot 2024-06-10 at 12 58 54

ChaconC (approved):
LGTM and all tests worked, great work!

I  did run into a case where the banner stayed on screen indefinitely, but that was caused by a popup from my password manager (never triggering window focus to change).

A minor comment about removing banners on success but feels ok to pre-approve.

src/shared/utils/hooks/useSessionBanners/useSessionBanners.ts
farmerpaul: The primary reason I opted to do it on logout (in the admin app, and then carried that functionality here), is to give the user confidence that their session is over and their state has been securely cleared (residual banners meant for them aren't bleeding into a now-anonymous session state).

I hear you with this particular banner persistence not making sense post-login success. I think it would make sense to just clear the specific error banner after login is successful, rather than do a blanket `removeAllBanners()` when a session is activated. IMO this same approach would make sense to do for _any_ form submission that at first may generate an error banner, then succeeds before the error banner has timed out. Thus I don't think it belongs in `useSessionBanners`. I'll make this change for the login screen for now.

This hasn't been identified in the admin app, but there may be situations where the same thing makes sense. We might start to see such things reported with more intensive QA.

farmerpaul: I made that change for both the login & and signup forms (which also exhibited this behaviour, so long as you're quick enough to submit the form successfully after an error is shown).

src/shared/ui/Banners/Banner/lib/types.ts
```diff
@@ -0,0 +1,10 @@
+import { AlertProps } from '@mui/material';
+
+export type BannerProps = {
+  /** @default 5000 */
+  duration?: number | null;
```
sultanofcardio, line 5: I think we could add here that `null` means persistent

sultanofcardio (approved):
Works as expected

src/pages/Login/index.tsx
farmerpaul: Indeed, that could be desirable in the future, however it would require a new strategy for identifying banners than we currently have in place. Could be that we provide an arbitrary unique `key` string, and separately provide the `*Banner` component that we wish to have rendered, rather than tightly couple the key with the component as we're doing now.

I'd say let's support that if we need it down the road.

PR #482, opened 2024-06-11
Log the admin out if there is a mismatched respondent ID

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-6844](https://mindlogger.atlassian.net/browse/M2-6844)

This PR updates the behaviour of the "mismatched respondent" take now validation scenario. The current behaviour shows an error message instructing the admin user to log out, log back in as the correct user, and re-initiate take now from the admin app

image

The updated behaviour automatically logs the admin out, and shows a persistent error message instructing them to log in as the correct user to initiate take now. The multi-informant take now context is retained in the tab as long as it remains open.

image

#### Before

image

#### After

image

### 🪤 Peer Testing

- Construct a take-now URL in the following format: `https://localhost:5173/protected/applets/{appletId}?startActivityOrFlow={activityId}&sourceSubjectId={sourceSubjectId}&targetSubjectId={targetSubjectId}&respondentId={randomUUID}`, where `randomUUID` is a random v4 UUID.
- Enter the URL in the browser and log in with valid credentials
- Observe and confirm:
  - The error message matches the expected message
  - The error message persists until you dismiss it (it is enough to confirm it lasts longer than 15s)
  - You are logged back out
- Log in again and confirm the same as above

### ✏️ Notes

There is a known defect that the error message will persist into a separate tab unless you dismiss it. A fix for this is coming in a separate PR

### DIFF
--- a/src/features/TakeNow/ui/ValidateTakeNowParams.tsx
+++ b/src/features/TakeNow/ui/ValidateTakeNowParams.tsx
@@ -2,9 +2,12 @@ import { useEffect } from 'react';
 
 import { appletModel } from '~/entities/applet';
 import { useBanners } from '~/entities/banner/model';
-import { TakeNowParams } from '~/features/TakeNow/lib/TakeNowParams.types';
+import { useLogout } from '~/features/Logout';
+import { TakeNowParams } from '~/features/TakeNow/lib/types';
 import { useTakeNowValidation } from '~/features/TakeNow/lib/useTakeNowValidation';
+import ROUTES from '~/shared/constants/routes';
 import Loader from '~/shared/ui/Loader';
+import { useCustomNavigation } from '~/shared/utils';
 import { ActivityGroups } from '~/widgets/ActivityGroups';
 
 function ValidateTakeNowParams({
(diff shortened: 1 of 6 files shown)


src/features/TakeNow/ui/ValidateTakeNowParams.tsx
```diff
@@ -44,10 +49,21 @@ function ValidateTakeNowParams({
...
+        duration = null;
+      } else if (error.type === 'invalidApplet') {
+        navigate(ROUTES.appletList.path);
+      }
+
+      setTimeout(() =>
+        addErrorBanner({
+          children: error.error,
+          duration,
+        }),
```
sultanofcardio, line 65: @farmerpaul I decided to split the banner fix into a separate PR because I haven't figured out an elegant solution for it yet

farmerpaul (approved):
Works well.

I did discover an unrelated bug, however: when you try to do Take Now for a mobile-only activity (which I now have thanks to having to test media items), you get some pretty jarring behaviour redirecting you to the Apple App Store – twice – without any indication as to why. Logged in JIRA [here](https://mindlogger.atlassian.net/browse/M2-6999).

PR #483, opened 2024-06-11
fix(mi): update ownerId for TakeNow (M2-6948)

ChaconC:
### 📝 Description

Replaces how we obtain the workspace ID (`owner_id`) to use the new value from `/applets/{applet_id}` instead of using `encryption.account_id`

🔗 [Jira Ticket M2-6948](https://mindlogger.atlassian.net/browse/M2-6948)

### 🪤 Peer Testing

Ensure your local backend is using this branch https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1402

- Initiate TakeNow from Admin
- Log in to web app
- TakeNow should complete correctly

### ✏️ Notes

Merging blocked until BE PR is merged

### DIFF
--- a/src/features/TakeNow/lib/useTakeNowValidation.tsx
+++ b/src/features/TakeNow/lib/useTakeNowValidation.tsx
@@ -71,7 +71,7 @@ export const useTakeNowValidation = ({
 
   useEffect(() => {
     if (!workspaceId && !isLoadingApplet && appletData) {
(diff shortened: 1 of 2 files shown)


sultanofcardio (approved):
I wasn't able to test this with the kind of data that would cause it to fail, but I confirmed that the property is present in the API response and has the correct value

PR #490, opened 2024-06-24
feat(mi): Web App Take Now event tracking (M2-6991)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-6991](https://mindlogger.atlassian.net/browse/M2-6991)

This PR tracks new/updated Mixpanel events related to taking an assessment in the Web App via Take Now.

### 🪤 Peer Testing

Please refer to [JIRA ticket](https://mindlogger.atlassian.net/browse/M2-6991) for details around when events are triggered and what event properties should be attached to those events.

#### Prerequisites:

- To test, you'll need to have access to the **MindLogger Analytics Dev** project in Mixpanel; please let @farmerpaul know if you are planning to do peer testing so that you can be granted access.
- Ensure you have these env vars set locally:
  ```
  VITE_ENV=dev
  VITE_MIXPANEL_TOKEN=c1e1e662f03d9f1363768731077ea000
  VITE_MIXPANEL_FORCE_ENABLE=true
  VITE_BUILD_VERSION=head
  ```

#### Testing:

* For each table row in the [JIRA ticket](https://mindlogger.atlassian.net/browse/M2-6991):
  1. Follow the instructions under the **Trigger** column to trigger the event.
  2. Open the **MindLogger Analytics Dev** project in Mixpanel, and navigate to the [Events tab](https://mixpanel.com/project/3333580/view/3841316/app/events). Refresh if needed.

      **Expected outcome:** The event matching the **Event Name** column should have been added to the top of the list.

  4. Expand the event and open the **Your Properties** tab:  
      Screenshot 2024-06-18 at 4 32 53 PM

      **Expected outcome:** The custom properties for that event mentioned in the **Details** column should be listed, as well as common events mentioned at the top of the JIRA ticket, which apply to all events.

> [!NOTE]
> Currently, the **Assessment Started** event is logged **twice in a row** when starting an assessment via Take Now (**four times** when `StrictMode` is enabled, which includes dev environments). It's unclear why this is happening, and hopefully @sultanofcardio can suggest a fix!

> [!NOTE]
> Secondly, when completing an **Activity Flow**, the **Assessment completed** event is logged once for each completed activity. This is existing analytics behaviour for the Web App; this PR just adds properties to these events. The `Activity Flow ID` should remain constant for each of those events, while the `Activity ID` property should reflect the activity just completed.

* Click the link under the Distinct ID column for any of these recently logged events.

    **Expected outcome:** Under the User Profile Properties box on the left, there should be a field called **App Build Number** having a value of `head`.


src/abstract/lib/types/multiInformant.ts
```diff
@@ -8,4 +8,8 @@ export type MultiInformantState = {
   currentUserSubject?: MultiInformantSubject;
   sourceSubject?: MultiInformantSubject;
   targetSubject?: MultiInformantSubject;
+  appletId?: string;
```
farmerpaul, line 11: Even though there are other ways to access current `appletId`, it was convenient to just include it here for consumption by `TakeNowSuccessModal`, which derives all its needed props from the most recent `MultiInformantState` stored in Redux.

sultanofcardio (commented):
Preliminary comments before testing. Currently looking into the re-render bug

src/widgets/ActivityGroups/ui/ActivityGroups.tsx
```diff
@@ -42,13 +49,18 @@ export const ActivityGroups = (props: Props) => {
...
+    ensureMultiInformantStateExists,
+  } = appletModel.hooks.useMultiInformantState();
 
   useOnceEffect(() => {
     ensureMultiInformantStateExists();
     if (featureFlags.enableMultiInformant) {
       if (isInMultiInformantFlow() && !props.startActivityOrFlow) {
+        setTakeNowSuccessModalState({ isOpen: true, ...getMultiInformantState() });
         resetMultiInformantState();
       }
```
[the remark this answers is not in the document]
farmerpaul: Thank you @sultanofcardio! It seemed a bit too good to be true for it to be that simple. 😅

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
```diff
@@ -101,20 +102,45 @@ export const ActivityCard = ({ activityListItem }: Props) => {
...
+  // Start activity on mount if direct linking to this activity; pass `true` to onStartActivity to
+  // ensure it doesn't resume from a previous progress state.
   useOnceEffect(() => {
     if (
       context.startActivityOrFlow &&
       ((!isFlow && context.startActivityOrFlow === activityListItem.activityId) ||
         (isFlow && context.startActivityOrFlow === activityListItem.flowId))
     ) {
-      restartActivity();
+      onStartActivity(true);
```
sultanofcardio, line 143: The real source to the mounting problem eludes me, but I wouldn't be surprised if it had to do with lack of memoization in a parent component because a lot of functions inside hooks are being wrapped in `useCalled`.

However, for now we can fix it by using a ref.


```typescript
const previouslyMounted = useRef(false);

// Other code...

useOnceEffect(() => {
    if (previouslyMounted.current) {
      return;
    } else {
      previouslyMounted.current = true;
    }
```

sultanofcardio: In fact, it may be useful to put this ref inside the definition of `useOnceEffect` 🤔

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
```diff
@@ -101,20 +102,45 @@ export const ActivityCard = ({ activityListItem }: Props) => {
...
+    }
+
+    Mixpanel.track(MixpanelEvents.ActivityRestarted, analyticsPayload);
   };
+
   const resumeActivity = () => {
     onStartActivity(false);
-    Mixpanel.track(MixEvents.ActivityResumed, { [MixProperties.AppletId]: context.applet.id });
+
+    const analyticsPayload: MixpanelPayload = {
```
sultanofcardio, line 121: We don't seem to do anything with this variable. Should it be removed? The old `ActivityResumed` event was only tracking applet ID

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
```diff
@@ -101,20 +102,45 @@ export const ActivityCard = ({ activityListItem }: Props) => {
...
+      [MixpanelProps.ActivityId]: activityListItem.activityId,
+    };
+
+    if (isFlow) {
+      analyticsPayload[MixpanelProps.ActivityFlowId] = activityListItem.flowId;
+    }
+
+    Mixpanel.track(MixpanelEvents.ActivityResumed, {
+      [MixpanelProps.AppletId]: context.applet.id,
+    });
```
sultanofcardio, line 132: Unfortunately, this `restartActivity` function is also called when the user manually clicks the start activity button on the activity card (outside take now).

https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/af848c2bc1fcca901d09e14b333c745ae6c3ad7a/src/widgets/ActivityGroups/ui/ActivityCard/ActivityCardRestartResume.tsx#L81-L90

Which causes a resume event to get created right after the started one

image

I think this is the existing behaviour, so it's totally fair if we want to ignore this. However, if we want to increase the fidelity of these events, we'll need to create a separate function that gets called for start activity

sultanofcardio (commented):
A couple other things I found. Everything else seems to work so once these are addressed, we can approve and merge

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
farmerpaul: I just missed passing this value to the `Mixpanel.track` call – good catch!

sultanofcardio: I was wondering if we should be tracking activity ID and Flow ID for the `ActivityResumed`, since they're not mentioned in the ticket and the old one wasn't tracking them. It makes sense of course, but I thought it was something to be noted

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
farmerpaul: So, I've tried this approach and it doesn't seem to work on my end. `previouslyMounted.current` is always evaluating as `false` and is never getting set to `true`, which means something else is going on.

I've also discovered more nuance around this:
- If the current activity being "Taken Now" was in a state of partial completion, the `useOnceEffect` only runs once.
- It's only when there is no saved activiy progress that the `useOnceEffect` runs twice.

I'll keep investigating, but feel free to contribute ideas!

sultanofcardio: Strict mode only runs useEffect twice in dev mode, so I decided to test the production build as well. I was able to reproduce the behaviour in both cases. I then updated the definition of `useOnceEffect` here: https://github.com/ChildMindInstitute/mindlogger-web-refactor/commit/b0c7e1c0453347b83f545c3824cb872a8b455404

I added a console log at the top of the `useOnceEffect` call, and another right above the call to `onStartActivity`

```typescript
useOnceEffect(() => {
  console.log(`Running ActivityCard.useOnceEffect (name: ${title})`);
  if (
    context.startActivityOrFlow &&
    ((!isFlow && context.startActivityOrFlow === activityListItem.activityId) ||
      (isFlow && context.startActivityOrFlow === activityListItem.flowId))
  ) {
    console.log(`Running onStartActivity (name: ${title})`);
    onStartActivity(true);
  }
});
```

And I got this output in each scenario

```txt
Running ActivityCard.useOnceEffect (name: Activity 1)
Running onStartActivity (name: Activity 1)
Running ActivityCard.useOnceEffect (name: Flow1)
Running ActivityCard.useOnceEffect (name: A/B Trails iPad)
Running ActivityCard.useOnceEffect (name: A/B Trails Mobile)
Running ActivityCard.useOnceEffect (name: Simple & Choice Reaction Time Task Builder)
Running ActivityCard.useOnceEffect (name: CST Gyroscope)
Running ActivityCard.useOnceEffect (name: CST Touch)
```

Give that a test and let me know if it still runs twice

farmerpaul: FWIW I've just been disabling `<StrictMode>` altogether in `Main.tsx` while troubleshooting this, so that shouldn't be a factor for me. I've made the changes exactly you outlined above.

So long as I keep the activity in a state of partial-completion, it does **not** run twice:

CleanShot 2024-06-25 at 13 51 48@2x

But if I make sure I've completed the activity first, then start Take Now, it still continues to run twice:

CleanShot 2024-06-25 at 13 51 22@2x

But on your end, with your change you're unable to reproduce the buggy behaviour for both scenarios?

farmerpaul: Another detail worth noting – I'm not logging out after completing the activity. It only happens if I have an active Web App login session, and I'm starting Take Now for an activity in a not-in-progress state.

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
farmerpaul: Fixed this! Thanks for noticing!

src/shared/utils/hooks/useOnceEffect.ts
```diff
@@ -1,11 +1,19 @@
...
+ * a ref and `useLayoutEffect` to ensure that the effect runs only once after the DOM has rendered.
+ *
  * @param effect Some code that you'd normally put in a `useEffect` hook.
  */
 export const useOnceEffect = (effect: EffectCallback) => {
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  useEffect(effect, []);
+  const hasExecutedRef = useRef(false);
+
+  useLayoutEffect(() => {
```
sultanofcardio, line 12: These aren't the same thing. They should be separate functions so that the consumer can use them instead of `useEffect` and `useLayoutEffect` as appropriate. Right now we only have one use case for `useOnceLayoutEffect`

farmerpaul: Oh weird, when I looked at your branch earlier I thought just changed the existing hook rather than introduced a new one. My bad, I must have been misreading it! I was a bit concerned about making a sweeping change as well but thought it was safe/reasonable, deferring to what I thought you'd pushed. I'll create a new hook then!

farmerpaul: Updated!

sultanofcardio (approved):

PR #499, opened 2024-07-01
(feat: M2-7146) [Web] Processing Screen implementation (Basic UI)

moiskillnadne:
### 📝 Description
🔗 [Jira Ticket M2-#](https://mindlogger.atlassian.net/browse/M2-#)

The processing screen implementation.

Changes include:

- [Thing]
- [Other thing]
- [More things]


src/features/PassSurvey/ui/SurveyHeader.tsx
```diff
@@ -25,6 +26,12 @@ const SurveyHeader = (props: Props) => {
     return str.length > length ? `${str.substring(0, length)}...` : str;
   };
 
+  const title = props.title ?? context.activity.name;
```
BamMironov, line 29: I wonder why it is supposed to be like this. When is the `context.activity.name` case used?

src/shared/constants/routes.ts
```diff
@@ -119,7 +119,7 @@ const ROUTES = {
       flowId: string | null;
       publicAppletKey: string | null;
     }) =>
-      `/protected/answer-processing?appletId=${appletId}&eventId=${eventId}&activityId=${activityId}&flowId=${flowId}&publicAppletKey=${publicAppletKey}`,
+      `/protected/answer-processing?appletId=${appletId}&eventId=${eventId}&activityId=${activityId}${flowId ? `flowId=${flowId}` : ''}${publicAppletKey ? `publicAppletKey=${publicAppletKey}` : ''}`,
```
BamMironov, line 122: Just a comment (no need to do it now): Maybe it would be a good idea to somehow utilize the [Tagged templated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates) to avoid multiple conditional statements in template strings.

src/widgets/SurveyAnswerProcessing/ui/ProcessingScreen.tsx
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  const onFinish = useCallback(() => {
+    const canBeClosed = true; // TODO: Change on real one when the store will be ready
+
+    if (!canBeClosed) {
+      return addWarningBanner(t('answerProcessingScreen.processInProgress'));
+    }
+
+    return context.flow ? completeFlow({ forceComplete: true }) : completeActivity();
```
BamMironov, line 35: How is it considered a "forced" complete action when `onFinish` is invoked by a user action (button press).

src/widgets/SurveyAnswerProcessing/ui/ProgressBar.tsx
```diff
@@ -0,0 +1,44 @@
...
+  activitiesCount,
+  activityName,
+  isCompleted,
+}: Props) => {
+  return (
+    <Box
+      display="flex"
+      borderRadius="12px"
+      padding="16px 8px"
+      marginTop="16px"
```
BamMironov, line 28: margins and paddings should be applied at the upper level, not inside the component

src/features/PassSurvey/ui/SurveyHeader.tsx
```diff
@@ -25,6 +26,12 @@ const SurveyHeader = (props: Props) => {
...
+
+  const isProgressDefined = props.progress !== undefined;
+
+  const titleMarginBottom = greaterThanSM ? '8px' : '10px';
+
   return (
     <Box
       paddingX={greaterThanSM ? '24px' : '16px'}
```
BamMironov, line 37: Please apply paddings in the outer (upper level) component. For instance, in SurveyLayout

BamMironov (changes requested):

src/features/PassSurvey/ui/SurveyHeader.tsx
moiskillnadne: Most of the time the title is used from the context.

moiskillnadne: The props title used only when you want to provide some specific title

src/shared/constants/routes.ts
moiskillnadne: Yeah, maybe it makes sense to think how I can improve the readability of this part of the code

src/widgets/SurveyAnswerProcessing/ui/ProgressBar.tsx
moiskillnadne: Done

src/widgets/SurveyAnswerProcessing/ui/ProcessingScreen.tsx
moiskillnadne: The "forceComplete" flag is provided to the method to navigate the user after flow completion correctly.
BTW: This part of code very likely will be reimplemented in the next PR

src/features/PassSurvey/ui/SurveyHeader.tsx
moiskillnadne: I can`t do it. These paddings is applied to the header specifically.

BamMironov (approved):

PR #506, opened 2024-07-17
feat: Updating subject fetching logic (M2-7035)

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-7035](https://mindlogger.atlassian.net/browse/M2-7035)

#### Background

In the take now flow, it is necessary to fetch the subject details for the following parties:
- Current authenticated user
- Source subject
- Target subject

It is currently only possible to fetch the subject details of the source and target subjects if you are a team member. When https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1494 is merged, full account participants will also be able to perform this operation when there exists a relation between them and the subject they wish to fetch the details of.

It is also currently only possible for the current user to fetch their own subject details if they are a team member.

#### What's changed

In this PR, I've modified the way subject fetching is done to make use of the new Get My Subject (`/users/me/subjects/{applet_id}`) endpoint, which allows both full account participants and team members to consume it.

I've also modified the way subject fetching is done to eliminate duplicate requests. The web app will now fetch the subject details in this order:
- current user
- source subject
- target subject

If any of the subsequent subject IDs match one of the former ones, the details are reused and the request is not sent. This is not just to cut down on unnecessary requests, it is a consequence of the need for a temporary relation in order for full account participants to fetch the subject details. In the admin panel (from https://github.com/ChildMindInstitute/mindlogger-admin/pull/1866), a temporary relation is only created when the subject IDs don't match, so it was necessary to prevent fetching the details in the same scenario.

### 📸 Screenshots

N/A

### 🪤 Peer Testing

> [!NOTE]
> If https://github.com/ChildMindInstitute/mindlogger-admin/pull/1866 has not yet been merged, test this PR locally using the `feature/M2-7035-participant-take-now` branch of the admin panel

- Initiate take now using the steps outlined in the testing section of that PR for all the scenarios
- Confirm that the take now flow is successfully started in each scenario

### ✏️ Notes

N/A


mbanting:
@sultanofcardio I tested this out. Unfortunately the BE (`develop`) is still returning 403s. 
It looks like it is calling the `/users/me/subjects/` as mentioned in the PR description, but if the respondent is also either the source or target, it looks up it's endpoint via the `/subjects/` where it then fails, assuming because of the following
> It is also currently only possible for the current user to fetch their own subject details if they are a team member.

Here's a recording: https://www.loom.com/share/e83cf4fa8d684fae9e12dfb0dc298b45?sid=007cf3ef-3679-4b31-8bc4-34d2e791348d

src/features/TakeNow/lib/useTakeNowValidation.tsx
```diff
@@ -23,7 +23,8 @@ export const useTakeNowValidation = ({
   respondentId,
   multiInformantAssessmentId,
 }: TakeNowParams): TakeNowValidatedState => {
-  const [workspaceId, setWorkspaceId] = useState<string | null>();
+  const [fetchSourceSubject, setFetchSourceSubject] = useState<boolean>();
+  const [fetchTargetSubject, setFetchTargetSubject] = useState<boolean>();
```
mbanting, line 27: Should these be initialized to false?

sultanofcardio: Thanks Marty. This is part of the bug you observed. I have some logic that depends on this being initially set to `undefined` so I've fixed disabling the query by explicitly comparing to `true`.

mbanting (approved):
Test and it is working now 
I am seeing a couple of Take Now Participant bugs but will create separate tickets for them as they don't seem to be related to this.
Code LGTM too. Great job @sultanofcardio

PR #521, opened 2024-08-08
(feat: M2-6790) Conditional logic improvements: Number select

moiskillnadne:
### 📝 Description
🔗 [Jira Ticket M2-6790](https://mindlogger.atlassian.net/browse/M2-6790)


src/features/PassSurvey/model/ConditionalLogicBuilder.ts
```diff
@@ -0,0 +1,100 @@
...
+import { appletModel } from '~/entities/applet/';
+import { Condition } from '~/shared/api';
+
+type Item = appletModel.ItemRecord;
+
+export type ItemMapByName = Record<string, Item>;
+
+class ConditionalLogicBuilder {
+  private hiddenOrSkippedItemNames: Set<string> = new Set();
+  private hiddenOrSkippedItemIds: Set<string> = new Set();
```
BamMironov, line 12: But how about creating an array like this:
```
private hiddenOrSkippedItems: Array<Item>
```

I understand the convenience of using Sets with primitives: it filters out duplicates automatically upon invoking "add". However, it populates the code with unnecessary lines likes this doing the same thing for the same objects:

this.hiddenOrSkippedItemNames.delete(item.name);
this.hiddenOrSkippedItemIds.delete(item.id);

src/widgets/Survey/model/hooks/useSurveyState.ts
```diff
@@ -1,16 +1,17 @@
...
   const items = useMemo(() => activityProgress?.items ?? [], [activityProgress.items]);
 
   const visibleItems = items.filter((x) => !x.isHidden);
 
-  const processedItems = appletModel.conditionalLogicBuilder.process(visibleItems);
+  const processedItems = PassSurveyModel.conditionalLogicBuilder.process(visibleItems);
 
   const conditionallyHiddenItemIds =
-    appletModel.conditionalLogicBuilder.getConditionallyHiddenItemIds();
+    PassSurveyModel.conditionalLogicBuilder.getConditionallyHiddenItemIds();
```
BamMironov, line 14: I believe, at this point, we don't even need a separate method to return hidden items. We could either:
1) just figure out the hidden items right here, in this function, as we have all items (visibleItems) and  processedItems.
2) or make the `process` method return both hidden and visible items:
```
const availableItems = items.filter((x) => !x.isHidden);
const { visibleItems, hiddenItems } = PassSurveyModel.conditionalLogicBuilder.process(availableItems);
```

BamMironov (changes requested):
Please consider changes below

src/features/PassSurvey/model/ConditionalLogicBuilder.ts
```diff
@@ -0,0 +1,100 @@
+import { ConditionalLogicValidator } from './ConditionalLogicValidator';
+
+import { appletModel } from '~/entities/applet/';
+import { Condition } from '~/shared/api';
+
+type Item = appletModel.ItemRecord;
+
+export type ItemMapByName = Record<string, Item>;
+
+class ConditionalLogicBuilder {
```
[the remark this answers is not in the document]
moiskillnadne: Done

src/features/PassSurvey/model/ConditionalLogicBuilder.ts
```diff
@@ -0,0 +1,100 @@
...
+  private hiddenOrSkippedItemNames: Set<string> = new Set();
+  private hiddenOrSkippedItemIds: Set<string> = new Set();
+
+  public process(items: Item[]): Item[] {
```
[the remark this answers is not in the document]
moiskillnadne: Done

src/features/PassSurvey/model/ConditionalLogicBuilder.ts
moiskillnadne: It doesn`t really help to reduce code. If we change a set to an array we should find the item each time when we want to manipulate it. 

So I don`t think that it improves readability or something. I suggest to keep it as is.

src/widgets/Survey/model/hooks/useSurveyState.ts
moiskillnadne: Done

src/features/PassSurvey/model/ConditionalLogicBuilder.ts
```diff
@@ -7,22 +7,32 @@ type Item = appletModel.ItemRecord;
 
 export type ItemMapByName = Record<string, Item>;
 
-class ConditionalLogicBuilder {
+type ProcessResult = {
+  visibleItems: Item[];
+  hiddenItemIds: Set<string>;
+};
+
+interface IConditionalLogicFilter {
```
BamMironov, line 15: I suggest renaming it to `IFilter`and moving to the shared level

BamMironov (changes requested):

src/features/PassSurvey/model/ConditionalLogicBuilder.ts
moiskillnadne: Done

src/shared/utils/types/filter.ts
```diff
@@ -0,0 +1,3 @@
+export interface IFilter<ItemProp, Result> {
+  filter: (items: ItemProp[]) => Result;
+}
```
BamMironov, line 3: I suggest changing it to this:
```
export interface IFilter<TItem, TResult = TItem[]> {
  filter(items: TItem[]): TResult;
}
```

1. Generic types are conventionally prefixed by "T"
2. `TResult` should be optional
3. Unless one wants to provide their own `TResult` type, the type should be the same `ItemProp[]`
4. I made it a class method, while yours is a property storing a function.

BamMironov (changes requested):

src/shared/utils/types/filter.ts
moiskillnadne: Done

BamMironov (approved):

PR #530, opened 2024-08-30
feat: MI Assign home screen (M2-7405)

farmerpaul:
- [x] Tests for the changes have been added
- [ ] Related documentation has been added / updated
- [ ] OSS packages added to MindLogger [open source credit page](https://mindlogger.atlassian.net/jira/servicedesk/projects/MLA/knowledge/articles/340623543?spaceKey=MLA)

### 📝 Description

🔗 [Jira Ticket M2-7405](https://mindlogger.atlassian.net/browse/M2-7405)

Adds support for displaying assigned activities and flows in the Web App, including:

- Displays new `TargetSubjectLabel` chip component on multi-informant assignments
- Lists multiple instances of `ActivityCard` for the same activity/flow when user has multiple assignments for that activity/flow
  - _Currently, an activity can both be set to `autoAssign: true` as well as have manual assignments, and the web app respects this possibility. In [a future Admin App ticket](https://mindlogger.atlassian.net/browse/M2-7604), UI restrictions will be imposed to make auto-assign and manual assignments mutually exclusive._
- Tracks distinct in-progress state of each activity/flow for each assignment
- For those activities/flows that are scheduled for one-time completion or on a daily schedule: Uses new `targetSubjectId` returned by the `/answers/applet/{applet_id}/completions` endpoint to compare  assignment has already been completed that day

Also fixes a couple missing translations and incorrect usage of i18next plural forms.

> [!TIP]
> Although many files were touched in this PR, many of those constituted tying `targetSubjectId` to activity progress in [this commit](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/530/commits/60f907138f10f802a25054f7c8aa037f5ad71995), and on the whole were somewhat repetitive. Strongly typing the `targetSubject` and `targetSubjectId` props to be required prop (cannot be `undefined`) made it easier to make that change in a comprehensive way. I attempted to group other related changes into their own commits and it may be easiest to review [commit-by-commit](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/530/commits).

### 📸 Screenshots

https://github.com/user-attachments/assets/6f7d37f6-ce17-45db-8654-06ab5dd50381

### 🪤 Peer Testing

> [!NOTE] 
[...]

PR #533, opened 2024-08-30
feture: Styling updates to paragraph and text items (M2-7458)

ramirlm:
🔗 [Jira Ticket M2-7458](https://mindlogger.atlassian.net/browse/M2-7458)

Design updates to both paragraph and text items, including:
- Border radius
- Error handling
- Error state styling

### 📸 Screenshots
CleanShot 2024-09-03 at 18 00 37

CleanShot 2024-09-03 at 18 03 04

### DIFF
--- a/src/entities/activity/ui/ActivityCardItem.tsx
+++ b/src/entities/activity/ui/ActivityCardItem.tsx
@@ -1,8 +1,8 @@
 import { useMemo } from 'react';
 
 import { AdditionalTextResponse } from './AdditionalTextResponse';
-import { Answer, hasAdditionalResponse, requiresAdditionalResponse } from '../lib';
 import { ItemPicker } from './items/ItemPicker';
+import { Answer, hasAdditionalResponse, requiresAdditionalResponse } from '../lib';
 
 import { appletModel } from '~/entities/applet';
 import { SliderAnimation } from '~/shared/animations';
--- a/src/shared/ui/Banners/SuccessBanner/index.tsx
+++ b/src/shared/ui/Banners/SuccessBanner/index.tsx
@@ -1,5 +1,5 @@
-import { Banner, BannerProps } from '../Banner';
 import { SUCCESS_BANNER_DURATION } from './lib/const';
+import { Banner, BannerProps } from '../Banner';
 
 export const SuccessBanner = ({ children, ...props }: BannerProps) => {
   return (
--- a/src/widgets/Survey/ui/index.tsx
+++ b/src/widgets/Survey/ui/index.tsx
@@ -94,7 +94,7 @@ export const SurveyWidget = (props: Props) => {
         publicAppletKey={publicAppletKey}
         appletId={appletId}
         errorLabel={
-          publicAppletKey ? t('additional.invalid_public_url') : (error?.evaluatedMessage ?? '')
+          publicAppletKey ? t('additional.invalid_public_url') : error?.evaluatedMessage ?? ''
         }
       />
     );


src/entities/activity/ui/items/TextItem.tsx
```diff
@@ -24,5 +35,24 @@ export const TextItem = ({ item, value, onValueChange, isDisabled }: TextItemPro
...
+      />
+      <Box
+        display="flex"
+        justifyContent="flex-end"
+        alignItems="center"
+        fontSize="small"
+        color={hasError ? `${Theme.colors.light.error}` : Theme.colors.light.outline}
+        mr={2}
+      >
+        {`${t('charactersCount', { numCharacters, maxCharacters: maxResponseLength })}`}
```
felipeMetaLab, line 54: force it as string is it necessary ? only { t(...)} would it return a string ?

src/shared/ui/Items/Text/index.tsx
```diff
@@ -35,18 +26,22 @@ export const TextItem = ({
...
                   width: '4px',
                 },
                 '&::-webkit-scrollbar-thumb': {
                   backgroundColor: '#C2C7CF',
-                  borderRadius: '100px',
                 },
               },
               '& .MuiInputBase-root': {
-                paddingRight: '2px', // Remove padding inside the input
+                paddingRight: '2px', // Reduce padding inside the input
```
felipeMetaLab, line 38: consider remove comments.

src/entities/activity/ui/items/TextItem.tsx
```diff
@@ -11,10 +17,15 @@ type TextItemProps = {
...
+  const [hasError, setHasError] = useState(false);
+
+  const numCharacters = value?.length || 0;
+
+  const { t } = useCustomTranslation();
 
   const onHandleValueChange = (value: string) => {
     if (value.length > maxResponseLength) {
-      return;
+      setHasError(true);
```
ChaconC, line 28: `setHasError` should happen regardless of the value (similar to ParagraphText)

CleanShot 2024-09-03 at 10 58 24@2x

src/entities/activity/ui/items/TextItem.tsx
```diff
@@ -11,10 +17,15 @@ type TextItemProps = {
 
 export const TextItem = ({ item, value, onValueChange, isDisabled }: TextItemProps) => {
   const { maxResponseLength } = item.config;
+  const [hasError, setHasError] = useState(false);
```
ChaconC, line 20: Should be initialized with the proper value
CleanShot 2024-09-03 at 10 57 52@2x

ChaconC (changes requested):

felipeMetaLab (commented):
consider these 2 reviews

src/entities/activity/ui/items/TextItem.tsx
ChaconC: It's showing 304/300 (error) in a default color variant (meaning no error). This was after I moved to the next question and then back

src/widgets/Survey/model/validateItem.ts
```diff
@@ -29,6 +29,17 @@ function isAnswerShouldBeNumeric(item: appletModel.ItemRecord) {
...
 
+function isAnswerTooLarge(item: appletModel.ItemRecord) {
+  if (
+    !['paragraphText', 'text'].includes(item.responseType) ||
+    !('maxResponseLength' in item.config) ||
+    typeof item.answer[0] !== 'string'
+  ) {
+    return false;
+  }
+  return item.answer[0].length > item.config.maxResponseLength ?? true;
```
ChaconC, line 40: Do you need this fallback value?

src/i18n/en/translation.json
```diff
@@ -19,6 +19,9 @@
     "pleaseProvideAdditionalText": "Please provide additional text",
     "pleaseListenToAudio": "Please listen to the audio until the end.",
     "onlyNumbersAllowed": "Only numbers are allowed",
+    "answerTooLarge": "Please provide a shorter answer",
+    "question_count_plural": "{{length}} questions",
+    "question_count_singular": "{{length}} question",
```
felipeMetaLab, line 24: I didn't find these question_count_singular, question_count_plural on fr translation, is it right ?

felipeMetaLab: oh I see, so question_count_plural and question_count_singular is not relevant to be added on fr Translation ?
I understand its a point that you only resolved the conflict, but I didnt find this on fr Translation anyway, if it is not relevant ok, we can proceed without add those variable to fr translation.

felipeMetaLab: no haha, I suggested to add on fr Translation, but I see, ok if you dont think its relevant, for me that's ok, I agree with Carlos review and you can resolve my review.

ChaconC (approved):

src/i18n/en/translation.json
ChaconC: Good catch @felipeMetaLab 

@ramirlm this is a side effect from your merge from dev into this branch. These lines were removed in [this PR](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/530/files#diff-748406503f0dd61aa7860bd36e02a6674aac6247b7e9bbf07dd262966fbaeff0) so they shouldn't be added back in here.

ChaconC:
Please review failed lint and test checks @ramirlm

ChaconC (approved):

PR #534, opened 2024-09-04
feat: MI Assign activity screens (M2-7406)

farmerpaul:
- [ ] Related documentation has been added / updated
- [ ] OSS packages added to MindLogger [open source credit page](https://mindlogger.atlassian.net/jira/servicedesk/projects/MLA/knowledge/articles/340623543?spaceKey=MLA)

### 📝 Description

🔗 [Jira Ticket M2-7406](https://mindlogger.atlassian.net/browse/M2-7406)

Adds informative banners and labels to Activity screens when performing a multi-informant assignment.

- Displays a new banner on the activity welcome screen informing the respondent who the assessment is about
- Displays a label within each item's description (`About {{name}}`) below any media, and above the first line containing text content

### 📸 Screenshots

| Before                      | After                                 |
| CleanShot 2024-09-04 at 17 57 47@2x |  CleanShot 2024-09-04 at 17 57 31@2x |
| CleanShot 2024-09-04 at 17 57 58@2x | CleanShot 2024-09-04 at 17 56 53@2x |
| CleanShot 2024-09-04 at 17 58 09@2x| CleanShot 2024-09-04 at 17 57 06@2x |

### 🪤 Peer Testing

> [!NOTE] 
> This feature depends on the `enableActivityAssign` feature flag. This should already be enabled in dev/local environments. It would also be worth testing this with that feature flag **disabled** to confirm that existing behaviour is preserved with the MI Assign feature disabled, by adding `features.enableActivityAssign = false` prior to the `return` statement [here in `useFeatureFlags`](https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/47b863f1ed86f17b5120e4754cb5d07a4b2fde4b/src/shared/utils/hooks/useFeatureFlags.ts#L16-L17).

Please refer to [ticket's AC](https://mindlogger.atlassian.net/browse/M2-7406) for the main peer testing steps.

In addition, after completing a multi-informant assigned activity, confirm that the POST to the `answers` endpoint includes the appropriate `targetSubjectId`. Also confirm that for self-reports, that property is not included.


farmerpaul:
I incorporated your suggestions, @sultanofcardio. Good catch on the interaction of Take Now with MI Assign. It's not just the UI issue you identified that was introduced with MI Assign – Take Now will actually no longer start unless you are assigned (manually or auto) to the activity as a self-report. Since Take Now should take precedence over MI Assign, however, this needs to be fixed. I've created [M2-7796](https://mindlogger.atlassian.net/browse/M2-7796) to address this.

sultanofcardio (approved):
Looks good

mbanting (approved):
Nice work @farmerpaul. Just one small comment but pre-approving.

Peer tested successfully aside from the badge and banner also showing on Take Now, as @sultanofcardio [pointed out](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/534#pullrequestreview-2290157259). The suggested solution in `ActivityCard/index.tsx` seems to work, although I'm not entirely sure I understand how? If going with it, I suggest adding a comment so others coming across it understand why `&& activityListItem.targetSubject === null` is needed.

farmerpaul:
@mbanting incorporated your feedback! Thanks for the 👀

PR #536, opened 2024-10-10
fix: Allow "Take Now" to start when respondent is not explicitly assigned (M2-7796)

JustinNusca:
### 📝 Description

🔗 [M2-7796](https://mindlogger.atlassian.net/browse/M2-7796): [R2][FE][MI Assign] Ensure Take Now takes priority over Assignments

This PR updates how "Take Now" activities/flows are started. Previously, this logic was handled via an effect hook rendered by the `ActivityCard` component, which was rendered as part of the list of activities/flows that appeared on page for the relevant applet.

The addition of activity/flow assignments (gated by the `enableActivityAssign` feature flag), changes the behaviour of this list so that only explicitly assigned activities/flows appear. 

This means that if Take Now was initiated for a respondent subject that did not have that the selected activity or flow assigned, it would fail to start. Additionally, if Take Now was initiated about a specific target subject, for a respondent subject that did not have that the selected activity or flow assigned _for that target subject_, it would fail to start.

This PR makes updates to change how "Take Now" is started, to address these scenarios.

I have done so my removing the hook from the `ActivityCard`, and adding a new hook, `useTakeNowRedirect`. This hook performs appropriate checks and will automatically call `startSurvey` (resulting in a navigation to the appropriate activity/flow route), when the appropriate parameters are provided.

This hook has been added to the `ActivityGroups` component, which is the highest-level component in the `AppletDetailsPage` that this hook could be easily placed in.

#### When Take Now Activity/Flow is unassigned

| Before | After |
|-|-|
| unassigned-before | unassigned-after |

#### When Take Now Activity/Flow's target subject _differs_ from assigned

| Before | After |
|-|-|
| diff-target-before | diff-target-after |

### 🪤 Peer Testing

The below notes have been copied from the Jira task for convenience — they describe the appropriate setup and relevant testing pretty clearly!

> #### Setup:
> * Create activities:
>   * Activity A: auto-assign is `true`
>   * Activity B: auto-assign is `false`
>   * Activity C: auto-assign is `false`
>   * Flow A: auto-assign is `true`
>   * Flow B: auto-assign is `false`
>   * Flow C: auto-assign is `false`
>
> * Ensure a User X exists on the applet, either a Full Account or Team Member
> * Ensure a Subject Y exists on the applet, a Limited Account
>
> * Set up assignments for respondent User X (either full account or team account):
>   * User X has no manual assignments for Activity A, Activity B, Flow A, and Flow B
>   * User X is assigned as a respondent on Subject Y for Activity C and for Flow C (and no self-report)
>
> #### Testing Steps:
> For each Activity (A, B, and C), and again for each Flow (A, B, and C):
>
> 1. From the Admin App, perform Take Now on that activity/flow with the User X as the source subject (respondent) and Subject Y as the target subject.
> 2. In the Web App, ensure that Take Now proceeds with the assessment, with the correct "ℹ️" tooltip UI showing in the upper-left corner, and no multi-informant assignment-related banners or badges should be rendered in the UI.

[M2-7796]: https://mindlogger.atlassian.net/browse/M2-7796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/widgets/ActivityGroups/model/hooks/useTakeNowRedirect.ts
+++ b/src/widgets/ActivityGroups/model/hooks/useTakeNowRedirect.ts
@@ -1,0 +1,87 @@
+import { useMemo } from 'react';
+
+import { openStoreLink } from '~/abstract/lib';
+import { isSupportedActivity } from '~/entities/activity';
+import { appletModel } from '~/entities/applet';
+import { useStartSurvey } from '~/features/PassSurvey';
+import { AppletBaseDTO, AppletEventsResponse, HydratedAssignmentDTO } from '~/shared/api';
+
+export function useTakeNowRedirect({
+  activityOrFlowId,
+  applet,
+  assignments = [],
+  events = [],
+  isPublic = false,
+  publicAppletKey = null,
+}: {
+  activityOrFlowId?: string | null;
+  applet?: AppletBaseDTO;
+  assignments?: HydratedAssignmentDTO[] | null;
+  events?: AppletEventsResponse['events'];
+  isPublic?: boolean;
+  publicAppletKey?: string | null;
+}) {
+  const { startSurvey } = useStartSurvey({ applet, isPublic, publicAppletKey });
+  const { getMultiInformantState } = appletModel.hooks.useMultiInformantState();
+
+  const handleTakeNowRedirect = useMemo(
+    () =>
+      ({
+        activityId,
+        eventId,
+        flowId = null,
+        isSupported = true,
+        targetSubjectId = null,
+      }: {
+        activityId: string;
+        eventId: string;
+        flowId?: string | null;
+        isSupported?: boolean;
+        targetSubjectId?: string | null;
+      }) => {
+        if (!isSupported) {
+          openStoreLink();
+          return;
+        }
+
+        startSurvey({ activityId, eventId, flowId, shouldRestart: true, targetSubjectId });
+      },
+    [startSurvey],
+  );
+
+  if (assignments && applet && events && activityOrFlowId) {
+    const activity = applet.activities.find(({ id }) => id === activityOrFlowId);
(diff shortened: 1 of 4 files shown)


src/widgets/ActivityGroups/model/hooks/useTakeNowRedirect.ts
```diff
@@ -0,0 +1,87 @@
...
+        flowId?: string | null;
+        isSupported?: boolean;
+        targetSubjectId?: string | null;
+      }) => {
+        if (!isSupported) {
+          openStoreLink();
+          return;
+        }
+
+        startSurvey({ activityId, eventId, flowId, shouldRestart: true, targetSubjectId });
```
farmerpaul, line 47: This is the culprit causing the "About …" labels to be shown in the assessment. This parameter was in fact added strictly for MI Assignment, not for Take Now, while Take Now uses a different mechanism for tracking target subject (and source subject) IDs.

The parameter name below could have been named more clearly (like `assignmentTargetSubjectId`), but alas!

So I would remove all references to `targetSubjectId` within this hook, and pass null here, and you should be good to go!

```suggestion
        startSurvey({ activityId, eventId, flowId, shouldRestart: true, targetSubjectId: null });
```

JustinNusca: Updated in [this commit](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/536/commits/3941a9fec1cd48ea74b54625377170cacc364ab6)!

farmerpaul: Tested and looks like it's working!

farmerpaul (approved):

PR #539, opened 2024-10-22
fix: Warning banner & post-login redirect both fail when starting `Take Now` while already logged into Web App as wrong respondent (M2-8029) 

LashaunnaS:
### 📝 Description

🔗 [Jira Ticket M2-8029](https://mindlogger.atlassian.net/browse/M2-8029)

- Fix to the  `useSessionBanners` hook to ensure that logic such as banner removal only occurs after authorization has changed. Replacing the `useEffect` with `useMemo` should ensure that any side effects that get triggered are due the state of the application once any further auth logic has been resolved.
- Fix to redirecting the user after correct authentication to navigate them to the 'Take Now' route.

### 📸 Screenshots

Before:

https://github.com/user-attachments/assets/8e5149f7-6291-4c96-956f-b24ef8f725c1

After:

https://github.com/user-attachments/assets/6102dd21-54d2-4405-9fde-5238354dabbe

### 🪤 Peer Testing

𝐏𝐫𝐞𝐜𝐨𝐧𝐝𝐢𝐭𝐢𝐨𝐧𝐬:
- Have an applet with at least one Team Member and one Full Account participant.
- Be already logged into the Web App as the Team Member.

𝐒𝐭𝐞𝐩𝐬 𝐭𝐨 𝐫𝐞𝐩𝐫𝐨𝐝𝐮𝐜𝐞:
- Perform Take Now on any activity, specifying the Full Account as the respondent, and anyone as the subject (self-report is otherwise).
- Observe the banner briefly shown at the top of the Web App when redirected to login screen.
- Log into Web App as the Full Account.
- Observe the next page you are led to in the Web App.

### DIFF
--- a/src/shared/utils/hooks/useSessionBanners/useSessionBanners.ts
+++ b/src/shared/utils/hooks/useSessionBanners/useSessionBanners.ts
@@ -1,4 +1,4 @@
-import { useMemo, useRef } from 'react';
+import { useRef } from 'react';
 
 import { useBanners } from '~/entities/banner/model';
 import { userModel } from '~/entities/user';
@@ -9,11 +9,9 @@ export const useSessionBanners = () => {
 
   const prevIsAuthorized = useRef(isAuthorized);
 
-  useMemo(() => {
-    if (prevIsAuthorized.current !== isAuthorized && !isAuthorized) {
-      removeAllBanners();
-    }
+  if (prevIsAuthorized.current !== isAuthorized && !isAuthorized) {
+    removeAllBanners();
+  }
 
-    prevIsAuthorized.current = isAuthorized;
-  }, [isAuthorized, removeAllBanners]);
+  prevIsAuthorized.current = isAuthorized;
 };
--- a/src/shared/utils/hooks/useSessionBanners/useSessionBanners.ts
+++ b/src/shared/utils/hooks/useSessionBanners/useSessionBanners.ts
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from 'react';
+import { useMemo, useRef } from 'react';
 
 import { useBanners } from '~/entities/banner/model';
 import { userModel } from '~/entities/user';
@@ -9,7 +9,7 @@ export const useSessionBanners = () => {
 
   const prevIsAuthorized = useRef(isAuthorized);
 
-  useEffect(() => {
+  useMemo(() => {
     if (prevIsAuthorized.current !== isAuthorized && !isAuthorized) {
       removeAllBanners();
     }
--- a/src/entities/user/model/hooks/useOnLogin.ts
+++ b/src/entities/user/model/hooks/useOnLogin.ts
@@ -41,7 +41,7 @@ export const useOnLogin = (params: Params) => {
     secureTokensStorage.setTokens(tokens);
 
     if (params.backRedirectPath !== undefined) {
-      navigate(params.backRedirectPath);
+      navigate(params.backRedirectPath, { replace: true });
     } else {
       Mixpanel.track('Login Successful');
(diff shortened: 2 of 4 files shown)


src/shared/utils/hooks/useSessionBanners/useSessionBanners.ts
```diff
@@ -9,7 +9,7 @@ export const useSessionBanners = () => {
 
   const prevIsAuthorized = useRef(isAuthorized);
 
-  useEffect(() => {
+  useMemo(() => {
```
farmerpaul, line 12: I can't say I actually understand what this change is doing, even after reading your PR description. I get that with a `useMemo`, the callback is executed inline immediately the first time it's encountered by the interpreter, whereas with `useEffect`, the callback is called async after the component has completed the initial render. Could you walk me through exactly what was going wrong here when `useEffect` was in place?

I've also never used `useMemo` for a situation where the return value was discarded, but maybe I just haven't stumbled upon this use-case before now.

The [React docs](https://react.dev/reference/react/useMemo#every-time-my-component-renders-the-calculation-in-usememo-re-runs) on `useMemo` say:
> **You should only rely on `useMemo` as a performance optimization.** If your code doesn’t work without it, find the underlying problem and fix it first. Then you may add useMemo to improve performance.

I guess I'm also wondering if this code aligns with that rule or not…

farmerpaul (commented):
Tested and works great! Thanks for adding the needed code to get `backRedirectPath` to actually be useful (seems it was a loose end that was never fully implemented prior to Metalab coming onto the project). Tested multiple scenarios, including trying to log in as the wrong user, and then finally, the right user, and Take Now was still triggered properly.

I just had one question about the `useMemo` that I hope you can help me understand!

src/shared/utils/hooks/useSessionBanners/useSessionBanners.ts
LashaunnaS: @farmerpaul that's a good callout. With further debugging, I think there may be a re-rendering issue stemming from the redux logic w/ `user`, `token`, and `isAuthorized` state. I can't quite pin down what's triggering the re-renders. 🤔

LashaunnaS: Removing the `useEffect` does ensure the banner isn't automatically cleared because of that particular component re-rendering. I do still have some concerns about the amount of re-rendering taking place, which seems to be a performance issue already established on prod.

I think the rerendering we're seeing is coming from how far along the app renders down the tree as the already authenticated user before it determines that the new user isn't authenticated.

farmerpaul (approved):
I'm glad it's working. I'd still like to get someone else's opinion here – @JustinNusca or @sultanofcardio, what do you think about [this change](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/539/files#diff-7a4944472491946803ff4d60e6dc6207a1ff8089132d63a9ae0855d80f811e6c)?

sultanofcardio (commented):
@farmerpaul thanks for the question. I think the particular change you referenced is fine - the `useEffect` didn't seem to be adding any value so the outcome is the same. In fact, it is what fixes the disappearing banner so great work ther @LashaunnaS.

This PR implements a general post-login redirection inside the `useLogout` hook, which will now apply in all cases of being logged out (not just take now). So if a user is idle while completing an activity and is logged out, it will result in them being redirected to that activity. I think this is a great addition, but is not strictly necessary for take now.

Post-login redirection in the take now scenario was never implemented. It just seems that way because the `multiInformantState` stays in redux as long as the tab remains open. After logging back in, the process for launching take now just starts over again using that existing state. This still works for me on the dev branch

PR #541, opened 2024-10-23
feat: implement content truncation for Action Plan cards (M2-7861)

qiushihe:
### 📝 Description

🔗 [Jira Ticket M2-7861](https://mindlogger.atlassian.net/browse/M2-7861)

This PR updates the Action Plan cards rendering code to implement a binary-searched based truncation algorithm. The algorithm will not only truncate at phrasal field level, but also at phrasal field content level, to ensure each card contains the maximum amount of content. Overflowing contents at the end of a card, and corresponding continuation content at the beginning of the following card, are connected by a set of ellipses.

For a walkthrough of implementation detail, please see this loom recording: https://www.loom.com/share/0f3fd7416f814e769ca1318e2ac13a28

### 📸 Screenshots

**Text Truncation Start**

Screenshot 2024-10-23 at 2 13 01 PM

---

**Text Truncation Both Ends**

Screenshot 2024-10-23 at 2 13 04 PM

---

**Text Truncation End**

Screenshot 2024-10-23 at 2 15 40 PM

---

**List Truncation Start/End **

Screenshot 2024-10-23 at 2 12 34 PM

---

**List Truncation Both Ends**

Screenshot 2024-10-23 at 2 13 15 PM

---

**List Truncation when Rendered as Sentence**

Screenshot 2024-10-23 at 2 14 34 PM

### 🪤 Peer Testing

> Note: By default, the maximum page height is `2504px`, which is very tall. So for testing, it's easier to run the code locally, then edit the `usePageMaxHeight` function of `src/entities/activity/ui/items/ActionPlan/hooks.tsx`, and make it return a smaller number (i.e. `512`).

1. Create an applet with an activity with a number of items
2. Create a phrasal template item with abundance of content
3. Go through the activity and reach the phrasal template step
4. The Action Plan cards should be truncated

### ✏️ Notes

Although the AC of the ticket only specified truncation of text content, the implementation actually also truncate list items. The reason being: if a list is too long, it should just be truncated anyway. Otherwise, we'll have to deal with all sort of complication around list content overflowing/cut-off, etc. If we REALLY don't want to truncate list items, we can still set a flag in the code to turn that off.

### DIFF
--- a/src/entities/activity/ui/items/ActionPlan/Phrase.tsx
+++ b/src/entities/activity/ui/items/ActionPlan/Phrase.tsx
@@ -1,27 +1,30 @@
-import React from 'react';
+import React, { ComponentProps, useMemo } from 'react';
 
 import Avatar from '@mui/material/Avatar';
 
-import {
-  useCorrelatedPageMaxHeightLineCount,
-  useXScaledDimension,
-  useYScaledDimension,
-} from './hooks';
-import { ActivitiesPhrasalData } from './phrasalData';
+import { DocumentData, PageComponent } from './Document.type';
+import { useXScaledDimension, useYScaledDimension } from './hooks';
 import { ResponseSegment } from './ResponseSegment';
 import { TextSegment } from './TextSegment';
 
-import { PhrasalTemplatePhrase, PhrasalTemplateField } from '~/entities/activity';
 import { Theme } from '~/shared/constants';
 import Box from '~/shared/ui/Box';
 
 export type PhraseProps = {
-  phrase: PhrasalTemplatePhrase;
-  phrasalData: ActivitiesPhrasalData;
-  noImage: boolean;
+  phraseId: string;
+  documentData: DocumentData;
+  pageComponents: PageComponent[];
+  isFirstOnPage: boolean;
+  isLastOnPage: boolean;
 };
 
-export const Phrase = ({ phrase, phrasalData, noImage }: PhraseProps) => {
+export const Phrase = ({
+  phraseId,
+  documentData,
+  pageComponents,
+  isFirstOnPage,
+  isLastOnPage,
+}: PhraseProps) => {
   const gap = useXScaledDimension(24);
   const minHeight = useXScaledDimension(72);
   const imageWidth = useXScaledDimension(67);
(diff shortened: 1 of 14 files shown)


src/shared/ui/Box/index.tsx
```diff
@@ -12,8 +12,8 @@ import MUIBox, { BoxProps } from '@mui/material/Box';
  * )
  */
 
-const Box = forwardRef<HTMLDivElement, BoxProps>((props: BoxProps) => {
-  return <MUIBox {...props} />;
+const Box = forwardRef<HTMLDivElement, BoxProps>((props, ref) => {
+  return <MUIBox {...props} ref={ref} />;
```
qiushihe, line 16: This doesn't really make or break anything. But technically the proper way to use `forwardRef` is to always define a 2-parameters function, with `ref` being the 2nd parameter. The old way also _works_ because `ref` is part of `...props`, but my IDE gave me a warning about the proper usage format, so I just changed it.

src/entities/activity/ui/items/ActionPlan/phrasalData.ts
```diff
@@ -77,7 +77,7 @@ export const extractActivitiesPhrasalData = (items: ItemRecord[]): ActivitiesPhr
         values: item.answer
           .map((value) => new Date(value))
           .filter((value) => !!value)
-          .map((value) => value.toLocaleTimeString()),
+          .map((value) => value.toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' })),
```
qiushihe, line 80: @JustinNusca this part is the fix for the time format issue.

src/entities/activity/ui/items/ActionPlan/pageComponent.ts
```diff
@@ -0,0 +1,296 @@
...
+    ctx: ActivityPhrasalDataSliderRowContext,
+    itemIndex: number,
+  ): FieldValueTransformer =>
+  (value) =>
+    t('sliderValue', { value, total: ctx.maxValues[itemIndex] });
+
+const joinWithComma: FieldValueItemsJoiner = (values) => values.join(', ');
+const joinWithDash: FieldValueItemsJoiner = (values) => values.join(' - ');
+
+export const buildPageComponents = (
```
qiushihe, line 45: I'm going to add some documentation for this function.

src/entities/activity/ui/items/ActionPlan/pageComponent.ts
```diff
@@ -0,0 +1,296 @@
...
+  return { imageUrlByPhraseId, hasImage };
+};
+
+export const getPagePhraseIds = (components: PageComponent[]) =>
+  components.reduce(
+    (acc, component) => (acc.includes(component.phraseId) ? acc : [...acc, component.phraseId]),
+    [] as string[],
+  );
+
+export const getFlatComponentIndices = (components: PageComponent[]): FlatComponentIndex[] =>
```
qiushihe, line 193: I'm going to add some documentation for this function.

src/entities/activity/ui/items/ActionPlan/pageComponent.ts
```diff
@@ -0,0 +1,296 @@
...
+  components.slice(0, inclusiveEnd + 1),
+  components.slice(inclusiveEnd + 1),
+];
+
+const divideTextItems = (items: string[], inclusiveEnd: number): [string[], string[]] => [
+  [...items.slice(0, inclusiveEnd + 1), '...'],
+  ['...', ...items.slice(inclusiveEnd + 1)],
+];
+
+export const deepDivideComponents = (
```
qiushihe, line 226: I'm going to add some documentation for this function.

src/entities/activity/ui/items/ActionPlan/Phrase.tsx
```diff
@@ -1,63 +1,98 @@
...
+          };
+        } else {
+          itemResponse = {
+            itemResponseType: 'text',
+            text: component.text,
+          };
+        }
+        rendered.push(<ResponseSegment itemResponse={itemResponse} isAtStart={isAtStart} />);
+      } else if (componentType === 'line_break') {
+        rendered.push(<br />);
```
JustinNusca, line 82: This came up in a discussion between @farmerpaul and I, for a [phrase-builder-related PR in the Admin app](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1958#discussion_r1816667345):

In Figma, the specs mention that the "line break" item should introduce [a 32px high gap](https://www.figma.com/design/3wcZ6bxK2QVo733ADxVjBw/SSI-2024?node-id=692-76427&t=BLw76cfZQzfaJAK6-4), which a `<br />` doesn't do. There also isn't a great cross-browser way of styling a `<br />` to include such a gap.

In my PR on the admin app, I render line breaks like so:

```jsx
<Box component="hr" sx={{ m: 0, height: 32, border: 'none' }} />
```

… which you could theoretically just drop in right here for the same effect.

This is a large PR that I haven't been able to look at in detail, so hopefully this doesn't have an adverse effect 😅.

qiushihe: I don't think it matter exactly what gets rendered, as the height calculation should just take whatever that gets rendered into account. So I should be able to swap this out.

qiushihe:
@farmerpaul the issue with the mobile browser vs. desktop browser truncation height difference is due to font-size and line-height calculation. Those 2 values are not set directly, but are calculated using useXScaledDimension and useYScaledDimension respectively. I can try tweaking the formula a bit, but at the end of the day, mobile browser is essentially just a responsive view, with difference page width than the desktop browser (which can also have whatever page width people end up with). So it unlikely that the 2 can be made to be always completely identical.

I also addressed those other comments, so if you can take another look that would be great. I'll just be tweaking the font-size/line-height stuff in the background for a bit.

farmerpaul (approved):
Hey @qiushihe, I recorded a Loom with a simple proposed solution to the aspect ratio issue, check it out: https://www.loom.com/share/898fea7503ef4f5696985e15eebf3fe4

See what you think, hopefully that will be adequate. I'll pre-approve regardless. 🙂

**Edit:** In my loom I demoed the solution with `useXScaledDimension`, but I supposed `useYScaledDimension` is actually the right one? Doesn't seem to make a whole lot of difference which one you use.

qiushihe:
That's actually really cool! I never thought about also scaling the max height value. And it seems easy enough, so I'll make the change. Thanks @farmerpaul!

src/entities/activity/ui/items/ActionPlan/hooks.tsx
```diff
@@ -29,11 +29,11 @@ export const usePageMinHeight = () => {
...
 
 export const usePageMaxHeight = () => {
-  // Use these for local development/testing when a shorter card would be
+  // Use these for local development/testing when shorter cards would be
   // easier to work with:
-  // return 512;
+  return useYScaledDimension(512);
 
-  return 2504;
+  // return useYScaledDimension(2504);
```
farmerpaul, line 36: Should the same change be made for `usePageMinHeight` above as well?

qiushihe: Yeah I suppose. I just also didn't think that number ever did anything. But I don't think it hurts to do it.

PR #542, opened 2024-10-24
feature: paragraph ItemResponseType for phrase builder response options WEB (M2-7762)

AlejandroCoronadoN:
- [ ] Added paragraphText phrasalTemplateCompatibleResponseTypes
- [ ] Assign item.reponseType with value paragraphText to the text-type rendering formatting on the ActionPlan. The entire paragraph should apper between fields of the ActionPlan. 

### 📝 Description
As an Admin, I want to choose the paragraph item type as a response phrase field so that I can include its contents in my Action Plan. By adding paragraph option to the allowed types on validate_phrasal_templates, the backend can store Activities with PhrasalBuilder Items that contain Parragraph items as elements of the PhraseBuilder item. 
 
🔗 [Jira Ticket M2-7762](https://mindlogger.atlassian.net/browse/M2-7762)

Changes include:
- [ ] The entire paragraph should apper between fields of the ActionPlan. 

### 📸 Screenshots

image

### 🪤 Peer Testing
* Create an activity
* Create and item  of type parragraph inside your activity
* In the same activity create a new item of type phrase builder
* Phrase builder should allow you to choose the newly created paragraph item inside the “Select Item (Response)” dropdown menu.
* Complete an activity from the web application and validate that the parragraph can be selected on the Phrase Builder section. 
* Invite a user to the new applet/activity
* Logged as a user complete the new applet/activity 
* On the last section of the Activity you will be able to see the new Action Plan with the paragraph inserted between the fields of the PhraseBuilder-ActionPlan-PhrasalTemplate

### DIFF
--- a/src/entities/activity/ui/items/ActionPlan/ResponseSegment.tsx
+++ b/src/entities/activity/ui/items/ActionPlan/ResponseSegment.tsx
@@ -24,7 +24,7 @@ const isAnswersSkipped = (answers: string[]): boolean => {
 type FieldValueTransformer = (value: string) => string;
 const identity: FieldValueTransformer = (value) => value;
 
-type FieldValuesJoiner = (values: string[]) => string;
+type FieldValuesJoiner = (values: string[]) => string | JSX.Element[];
 const joinWithComma: FieldValuesJoiner = (values) => values.join(', ');
 
 type ResponseSegmentProps = {
@@ -58,6 +58,9 @@ export const ResponseSegment = ({ phrasalData, field, isAtStart }: ResponseSegme
     };
   } else if (fieldPhrasalData.context.itemResponseType === 'timeRange') {
     joinSentenceWords = (values) => values.join(' - ');
+  } else if (fieldPhrasalData.context.itemResponseType === 'paragraphText') {
+    joinSentenceWords = (values) =>
+      values.map((item, index) => <div key={index}>{item || ' '}</div>);
   }
 
   let words: string[];
@@ -69,10 +72,8 @@ export const ResponseSegment = ({ phrasalData, field, isAtStart }: ResponseSegme
     words = isAnswersSkipped(fieldPhrasalData.values)
       ? [t('questionSkipped')]
       : fieldPhrasalData.values
-          .flatMap((value) => value.split(/\r?\n+/)) // Split each paragraph by newlines
-          .map((sentence) => sentence.trim()) // Trim whitespace around each sentence
-          .filter((sentence) => sentence.length > 0) // Filter out any empty strings
-          .map(transformValue); // Transform each sentence individually
+          .flatMap((value) => value.split(/\r?\n+/))
+          .map((sentence) => sentence.trim());
   } else if (fieldPhrasalDataType === 'indexed-array') {
     const indexedAnswers = fieldPhrasalData.values[field.itemIndex] || [];
     words = isAnswersSkipped(indexedAnswers)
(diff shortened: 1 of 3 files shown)


farmerpaul (changes requested):
Hey @AlejandroCoronadoN, I found that one of the bullets in AC wasn't satisfied:

> - An action plan configured with a phrase containing Paragraph Text item types will display the paragraph text, **including line breaks / new lines**

When I provided a paragraph of text with newlines in it, the newlines were removed in the Action Plan:

| Paragraph input | Action Plan |
| - | - |
| CleanShot 2024-10-29 at 16 30 44@2x | CleanShot 2024-10-29 at 16 30 54@2x |

AlejandroCoronadoN:
Hello @farmerpaul . Here are more details about my solution:
1. I added the conditional logic for paragraph type on fieldPhrasalDataType === paragraph
2. I splited my text into multiple "words" using breaklines so each "word" that is an individual sentence will be rendered in a separate <span> 
3. I added the conditional rendering for fieldPhrasalDataType === 'paragraph' inside the Text component. Here I rendered all the individual sentences in a different span. That way we can se a breakline between each paragraph. 

image
So this renders like this:
image


Following up with my comments on slack, I think there is another approach that involves rendering conditional on the fieldDisplayMode value. For this implementation I will need to:
1. Add a new fieldDisplayMode on the Admin application.
2. When creating the items I will need to insert the new displayMode for paragraphs
4. Using the NEW fieldDisplayMode I will render paragraphs differently. So the second condition where I use fieldPhrasalDataType === 'paragraph'  will be replaced by fieldDisplayMode === 'paragraph'. Making the code cleaner.

src/entities/activity/ui/items/ActionPlan/ResponseSegment.tsx
```diff
@@ -118,32 +126,52 @@ export const ResponseSegment = ({ phrasalData, field, isAtStart }: ResponseSegme
...
+                marginTop={isAtStart ? `0px` : undefined}
+                paddingLeft={`${listPadding}px`}
+              >
+                {words.map((item, index) => (
+                  <li key={index}>{item}</li>
+                ))}
+              </Box>
+            </>
+          );
+        } else if (fieldPhrasalDataType === 'paragraph') {
```
farmerpaul, line 154: I see why you created a closure here, because of the extra condition and wanting to avoid nested ternaries. However, you don't need to go to this trouble. It looks like me like Billie created the `joinSentenceWords` function as a variable for the very purpose of customizing it to do different things based on different item response types. So it would be easier to read, and a smaller change, if you just did this:

[Line 61](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/542/files#diff-ab9d6cf61610fc5f6da132ec966f87b80d8bf58ed8bdcbd1c60c17af110eec3dL61):
```tsx
  } else if (fieldPhrasalData.context.itemResponseType === 'paragraphText') {
    joinSentenceWords = (values) =>
      values.map((item, index) => <div key={index}>{item || ' '}</div>);
  }
```
(And to repeat from my last comment, the `' '` above is a non-breaking space.)

Then you can revert [this entire change (lines 121-146)](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/542/files#diff-ab9d6cf61610fc5f6da132ec966f87b80d8bf58ed8bdcbd1c60c17af110eec3dR129-R174).

src/entities/activity/ui/items/ActionPlan/ResponseSegment.tsx
```diff
@@ -65,6 +65,14 @@ export const ResponseSegment = ({ phrasalData, field, isAtStart }: ResponseSegme
...
       ? [t('questionSkipped')]
       : fieldPhrasalData.values.map(transformValue);
+  } else if (fieldPhrasalDataType === 'paragraph') {
+    words = isAnswersSkipped(fieldPhrasalData.values)
+      ? [t('questionSkipped')]
+      : fieldPhrasalData.values
+          .flatMap((value) => value.split(/\r?\n+/)) // Split each paragraph by newlines
+          .map((sentence) => sentence.trim()) // Trim whitespace around each sentence
+          .filter((sentence) => sentence.length > 0) // Filter out any empty strings
+          .map(transformValue); // Transform each sentence individually
```
farmerpaul, line 75: This code seems to filter out empty newlines… but I don't see why that would be desirable, and that's not actually asked for in the AC. I know that folks often insert 2 newlines in a row for the very purpose of visually delineating separate paragraphs. For instance, if I respond to a paragraph item with this response:

CleanShot 2024-10-31 at 18 11 20@2x

Then I, as the respondent, would expect it to look like this in the Action Plan:

CleanShot 2024-10-31 at 18 13 22@2x

However, with the empty lines filtered out, it ends up looking like this:

CleanShot 2024-10-31 at 18 13 54@2x

So to make sure the empty newlines are preserved, this code can be simplified to:

```suggestion
          .flatMap((value) => value.split(/\r?\n/)) // Split each paragraph by newlines
          .map(transformValue);
```

farmerpaul (commented):
Hey @AlejandroCoronadoN, thanks for making that change, it's almost there! I just noticed a couple ways the code could be made a bit clearer/simpler, and also wanted to draw your attention to a detail about consecutive line breaks.

Also, thank you for introducing me to `flatMap()`! I can't believe I've never really encountered that before… 🙈🙈🙈

src/entities/activity/ui/items/ActionPlan/ResponseSegment.tsx
AlejandroCoronadoN: I see. I thought we already had line breaks within the admin application, but you're right. I was only considering the admin side, assuming that if they needed line breaks, they could use the admin feature. However, this isn't the case for users filling out the activity. I'm working on these changes now.

src/entities/activity/ui/items/ActionPlan/ResponseSegment.tsx
```diff
@@ -24,7 +24,7 @@ const isAnswersSkipped = (answers: string[]): boolean => {
 type FieldValueTransformer = (value: string) => string;
 const identity: FieldValueTransformer = (value) => value;
 
-type FieldValuesJoiner = (values: string[]) => string;
+type FieldValuesJoiner = (values: string[]) => string | JSX.Element[];
```
farmerpaul, line 27: Nit: I often use the `ReactNode` type to represent "any rendered JSX content", it's a very flexible and convenient type to use:
```suggestion
type FieldValuesJoiner = (values: string[]) => ReactNode;
```

src/entities/activity/ui/items/ActionPlan/phrasalData.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
```
farmerpaul, line 1: Was this meant to be committed? I don't encounter any new prettier errors or warnings with this line removed:
```suggestion
```

farmerpaul (approved):
Looks good, thank you for making those changes! Just made one very minor type suggestion, and identified a rogue `eslint-disable` line that I wasn't sure was intentional. Anyway, PR is ready for QA 🙂 🙌🏻

PR #548, opened 2024-11-04
feat: Add web app activity group empty states (M2-7339)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-7339](https://mindlogger.atlassian.net/browse/M2-7339)

Add empty states to the activities list in the Web App.

Also fixed some minor spacing/styling surrounding activity list groups and the Restart button.

### 📸 Screenshots

**No activities available:**

CleanShot 2024-11-04 at 15 14 38@2x

**Only scheduled activities available:**

CleanShot 2024-11-04 at 15 13 50@2x

### 🪤 Peer Testing

1. Create an applet with only 1 activity whose auto-assign setting is **turned off** (so no assignments present).
2. Log into the Web App and open that applet.
    **Expected outcome:** The full-screen empty state should be shown
3. Create a new activity in that applet with auto-assign **turned on**, and schedule that activity to be available today sometime after the current time.
4. Log into the Web App and open that applet.
    **Expected outcome:** The "only scheduled activities available" empty state should be shown


src/widgets/ActivityGroups/ui/ActivityCard/ActivityCardRestartResume.tsx
```diff
@@ -83,15 +83,14 @@ export const ActivityCardRestartResume = ({
               padding: '10px 10px',
               transition: 'all 0.2s',
               minWidth: '120px',
+              mt: 2,
```
LashaunnaS, line 86: Do we have a decided pattern for defining styles as short form vs the long form "mt" vs "marginTop"?

src/widgets/ActivityGroups/ui/ActivityCard/ActivityCardRestartResume.tsx
```diff
@@ -83,15 +83,14 @@ export const ActivityCardRestartResume = ({
...
             }}
           >
-            <Box display="flex" flex={1} flexDirection="row" mt={2}>
-              <img src={ActivityRestartIcon} alt="Activity Restart Icon" />
+            <img src={ActivityRestartIcon} alt="Activity Restart Icon" />
```
LashaunnaS, line 89: Do we consider i18n in out alt tags?

LashaunnaS (approved):
Other than a few questions, this looks good.

src/widgets/ActivityGroups/ui/ActivityCard/ActivityCardRestartResume.tsx
farmerpaul: Good callout! I was just moving the `mt={2}` from below into here. But in fact, we shouldn't currently use the shorthand becuase the Material UI theme in the web was never configured properly, and I observed the shorthand values are being converted using an incorrect multiplier (something other than 10px, which is what's done in the Admin App).

The Web App was written with almost no attention given to styling best practices – it's kind of a mess. 😭

src/widgets/ActivityGroups/ui/ActivityCard/ActivityCardRestartResume.tsx
farmerpaul: Another great observation! Yes, this should be translated, I'll add it. I've seen i18n being neglected elsewhere in the web app. 😭😭

sultanofcardio (approved):

PR #559, opened 2024-12-05
chore: add Greek translation for empty state (M2-8133)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-8133](https://mindlogger.atlassian.net/browse/M2-8133)

Adds the Greek text for the "no activities" empty state.

### 📸 Screenshots

CleanShot 2024-12-05 at 14 28 40@2x
CleanShot 2024-12-05 at 14 29 06@2x

### 🪤 Peer Testing

1. Create an applet for which all activities are not assigned (auto-assign is off, no manual assignments).
2. Open the web app, set the language to Greek, navigate to the applet.
    **Expected outcome:** Empty state should show in Greek.


src/i18n/el/translation.json
```diff
@@ -329,6 +329,7 @@
       "support": "Υποστήριξη",
       "privacy": "Απόρρητο",
       "termsOfService": "Όροι"
-    }
+    },
+    "noActivities": "Δεν υπάρχουν διαθέσιμες δραστηριότητες για να ολοκληρώσετε αυτήν τη στιγμή"
```
ChaconC, line 333: I think this is meant to be `noApplets`, right? @ramirlm already had this in review https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/554

farmerpaul: No, this is a different ticket, actually. This is the empty state for an existing applet, but all activities are unassigned.

PR #562, opened 2024-12-17
fix: Fixing word break for large description option answers (M2-8267)

ramirlm:
### 📝 Description

🔗 [Jira Ticket M2-8267](https://mindlogger.atlassian.net/browse/M2-8267)

Changes include:

- Small CSS fix to prevent non-word wrap

### 📸 Screenshots

| Before (Optional)                      | After                                 |
| CleanShot 2024-12-17 at 10 51 48@2x | Code 2024-12-17 10 48 36 |

### 🪤 Peer Testing

On the admin panel:
- Create an applet with an activity with a large Option text item description
- Assign the activity

On the Web:
- Log in and start the activity
- Check if the item breaks the long description correctly(in words).


ChaconC (dismissed):
Tested with Single Select and Multi Select 👍

src/shared/ui/Items/SelectBase/SelectBaseText.tsx
```diff
@@ -15,10 +15,10 @@ export const SelectBaseText = (props: Props) => {
       lineHeight="28px"
       testid="select-text"
       sx={{
+        wordBreak: 'break-word',
         cursor: 'pointer',
-        lineBreak: 'anywhere',
+        hyphens: 'auto',
```
mbanting, line 20: This will still allow the browser is automatically break words at appropriate hyphenation points, following whatever rules it chooses. If the intent is to avoid breaking mid-word altogether, then we should use `[manual](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#manual)` or `[none](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#none)` instead.

mbanting (changes requested):

ChaconC (commented):
Thanks for the rework on this issue @ramirlm. I'm only seeing issues with the 45-character test cases, but this appears to be limited to mobile resolutions and caused by the title component. It might be worth mentioning this in the linked ticket.

CleanShot 2025-01-16 at 22 39 03

ChaconC (approved):

mbanting (approved):

PR #573, opened 2025-01-30
fix: Update Time/TimeRange/Date validations (M2-8302)

ChaconC:

### 📝 Description

Updates Time/TimeRange/Date item conditional logic validations

🔗 [Jira Ticket M2-8302](https://mindlogger.atlassian.net/browse/M2-8302)
🔗 [Jira Ticket M2-8619](https://mindlogger.atlassian.net/browse/M2-8619)

Changes include:

- Update mismatched TimeRange types
- Update time parse on Greater Than/Less Than/Equal To/Not Equal conditions
- Update date parse on Greater Than/Less Than/Equal To/Not Equal conditions
- Fixes crash when entering TimeRange values

#### Applet on Amplify
Use the provided applet in the linked amplify environment:
```
Credentials:
user: pr573@example.com
pass: pr573!
```
CleanShot 2025-01-30 at 15 13 41

- Complete the provided Activities for testing Time, Time Range and Date item type conditional logic
- Each activity includes a summary of test cases and expected validations
- Testing the same Applet on the `dev` environment will fail validations or crash

> [!NOTE]
> On "Between of" and "Outside of" conditions, boundaries are treated as failures.
> For example, for the rule "Time is **Between** **09:00** and **10:00**" a response of 09:00 will not pass validation.

#### Local Testing

- Create an applet containing Time, TimeRange and Date item types
- Add at least 1 other items
- Create conditional logic using: GREATER_THAN, LESS_THAN, EQUAL, NOT_EQUAL, BETWEEN, OUTSIDE conditional rules for each item

> [!NOTE]
> Use https://github.com/ChildMindInstitute/mindlogger-admin/pull/2023 to avoid issues while saving conditional logic Dates/Time item types

### DIFF
--- a/src/features/PassSurvey/model/ConditionalLogicValidator.test.ts
+++ b/src/features/PassSurvey/model/ConditionalLogicValidator.test.ts
@@ -1227,10 +1227,7 @@ describe('ConditionalLogicValidator', () => {
   });
 
   it('should validate EQUAL_TO_TIME correctly -> the times match', () => {
-    const time = {
-      hours: 10,
-      minutes: 30,
-    };
+    const time = '10:30';
 
     const condition: EqualToTimeCondition = {
       itemName: 'random-item-name',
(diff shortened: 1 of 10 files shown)


farmerpaul (approved):
Seeing these bugs fixed, thanks for preparing the demo activity! Code looks good.

PR #576, opened 2025-02-10
feat: Greek font update (M2-8442)

andrevitalb:
### 📝 Description

🔗 [Jira Ticket M2-8442](https://mindlogger.atlassian.net/browse/M2-8442)

This PR adds Lato as a font for greek characters with the use of CSS's `unicode-range` property. Atkinson remains as the default font for non-greek characters.

Some minor cleanup work was done as well for manual overrides along the project by defaulting MUI and the project to the following: `font-family: "Lato", "Atkinson", sans-serif` (configuration for the expected behavior added).

### 📸 Screenshots

| Before (Optional)                                                                         | After                                                                                     |
| ----------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------- |
|  |  |
|   |  | 

### 🪤 Peer Testing

This is only visually testable. Lato should be the default font for greek characters, Atkinson should be present elsewhere.

### ✏️ Notes

In order to achieve maximum cohesion between the two fonts used, Lato Black was used, instead of the default Lato Bold, since its weight more closely aligns with Atkinson's Bold.


farmerpaul (approved):
Thanks for cleaning up all the stray font definitions (there was basically no thought put into consolidating styling when this app was built 😞). Great work!

ChaconC (approved):

mbanting (approved):

mbanting (approved):

farmerpaul (approved):

PR #587, opened 2025-02-25
feat: Submit `eventHistoryId` in answer data (M2-8480)

farmerpaul:
### 📝 Description

> [!IMPORTANT]
> For the answer submission to be accepted by the BE, it requires running it with this branch checked out: https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1754

🔗 [Jira Ticket M2-8480](https://mindlogger.atlassian.net/browse/M2-8480)

This change adds the `eventHistoryId` property to answer submissions.

### 🪤 Peer Testing

1. Create an applet with an activity and an activity flow.
2. Log into the Web App.
3. Open DevTools Network tab.
4. Submit an assessment and observe the payload of the `POST /answers` request:
    **Expected outcome:** The JSON object should contain the property `eventHistoryId` with a value in this format: `[event ID]_[version]`.
5. While in DevTools, inspect the most recent result of the `GET /users/me/events/{applet_id}` endpoint and take note of the `version` associated with each event.
6. **Restart the activity**, proceed past the welcome screen, then click **Save & Exit**.
7. **Restart the flow**, proceed past the welcome screen, then click **Save & Exit**.
8. From the Admin App's **Applet > Schedule** tab, edit the event associated with each activity & flow and enable **Scheduled access** for each of them.
9. Reload the Web App, and again inspect the most recent result of the `GET /users/me/events/{applet_id}` endpoint, taking note of the `version` associated with each event.
    **Expected outcome:** They should be **incremented** from what you observed in step 5.
11. **Resume the activity**, submit the assessment, and observe the payload of the `POST /answers` request:
    **Expected outcome:** The JSON object should contain the property `eventHistoryId` with a value in this format: `[event ID]_[version]`, where `[version]` corresponds to the activity's event version found in **step 5**.
12. **Resume the flow**, submit each activity, and observe the payload of each `POST /answers` request associated with its activities:
    **Expected outcome:** The JSON object for each submission should contain the property `eventHistoryId` with a value in this format: `[event ID]_[version]`, where `[version]` corresponds to the flow's event version found in **step 5**.
13. Repeat steps 6-11, but click **Restart** instead of resuming.
    **Expected outcome:** The version found in the `eventHistoryId` property for each submission should correspond to the respective version observed in **step 5**.

### DIFF
--- a/src/features/PassSurvey/model/AnswersConstructService.ts
+++ b/src/features/PassSurvey/model/AnswersConstructService.ts
@@ -142,6 +142,7 @@ export default class AnswersConstructService implements ICompletionConstructServ
         width: window.innerWidth,
         height: window.innerHeight,
       },
+      eventHistoryId: `${this.event.id}_${this.event.version}`,
     };
   }
 
--- a/src/features/PassSurvey/model/AnswersConstructService.ts
+++ b/src/features/PassSurvey/model/AnswersConstructService.ts
@@ -142,7 +142,7 @@ export default class AnswersConstructService implements ICompletionConstructServ
         width: window.innerWidth,
         height: window.innerHeight,
       },
-      eventHistoryId: `${this.event.id}_${this.event.version}`,
+      ...(this.event.version && { eventHistoryId: `${this.event.id}_${this.event.version}` }),
     };
   }
 
--- a/src/abstract/lib/GroupBuilder/ActivityGroupsBuilder.test.ts
+++ b/src/abstract/lib/GroupBuilder/ActivityGroupsBuilder.test.ts
@@ -20,6 +20,7 @@ import { ActivityPipelineType, FlowProgress, GroupProgressState } from '~/abstra
 import { AvailabilityLabelType, EventAvailability } from '~/entities/event';
 import { PeriodicityType } from '~/entities/event';
 import { MIDNIGHT_DATE } from '~/shared/constants';
+import { mockEvent } from '~/test/utils/mock';
 
 const getProgress = (startAt: Date, endAt: Date | null) => {
   const result: GroupProgressState = {
@@ -30,6 +31,7 @@ const getProgress = (startAt: Date, endAt: Date | null) => {
       context: {
         summaryData: {},
       },
+      event: mockEvent,
     },
   };
   return result;
@@ -1580,6 +1582,7 @@ describe('ActivityGroupsBuilder', () => {
           context: {
             summaryData: {},
           },
+          event: mockEvent,
         },
       };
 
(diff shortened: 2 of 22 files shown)


src/features/PassSurvey/model/AnswersConstructService.ts
```diff
@@ -135,6 +135,7 @@ export default class AnswersConstructService implements ICompletionConstructServ
         width: window.innerWidth,
         height: window.innerHeight,
       },
+      eventHistoryId: `${this.event.id}_${this.event.version}`,
```
sultanofcardio, line 138: The chance of this happening may be unlikely, but just to be safe we should probably make sure both of these are here before including this property
```suggestion
      ...(this.event.version && { eventHistoryId: `${this.event.id}_${this.event.version}` }),
```

sultanofcardio (approved):
Looking good with one suggested change

ChaconC (approved):
Validated with updated BE, working now!

sultanofcardio:
@farmerpaul it shouldn't be long now, but be sure that https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1754 is merged before this PR

farmerpaul:
@sultanofcardio I've fixed this PR to align the event version associated with in-progress activities and flows with the event version that was available when the activity/flow was started. I think there are 3 new commits you can check out, some explanation found in the commit messages.

These changes should not change the behaviour that I brought up during standup of multiple activities/flows being listed for multiple overlapping events, as Natalia confirmed this is expected behaviour (though I'm going to follow up next week about the inconsistent timer settings issue, as there's a bit of confusion there – though it shouldn't affect this PR).

(@ChaconC feel free to ignore this one and continue focusing on Kenroy's PR, just keeping you in the loop!)

sultanofcardio:
@farmerpaul before you made the latest changes, I found that after completing step 12 of the test, if I made a new answer submission from scratch (i.e. not from an in-progress activity), it would still be using a previous event version number.

For example, during step 12 the in-progress activity was using event version `20250317-3`. Then, after making the changes in step 8 and refreshing the web app, the event would be at version `20250317-4`. Resuming/restarting the in-progress activity used version `20250317-3` as part of the `eventHistoryId` as expected. However, a subsequent submission of the activity still uses `20250317-3` instead of `20250317-4` (even though I can see this being returned from the BE in the network tab).

I haven't tested again since your latest change so this behaviour may no longer be present. If I test again later and find that it's still there, I'll try to discern the cause

farmerpaul:
@sultanofcardio, I believe I've reproduced. I guess I hadn't tested this scenario because I keep choosing a "Scheduled Access" option, which only allows one completion per day, so once it's been completed the activity is hidden. Instead, I updated the event by editing only the Timer settings, and was able to reproduce. I'll look into it!

farmerpaul:
@sultanofcardio fixed yet again! Fourth time's a charm??

Turns out an activity/flow's `GroupProgress` state is not deleted when the activity/flow is completed. It's just updated, and eventually it's reused when you start the activity/flow a second time (but has most of its properties removed – but notably, keeping the `event`, so that the "Restart" functionality works as desired and in that context).

There were a couple ways to I could have potentially solved this, but removing the `GroupProgress` entry seemed risky as I'm not entirely sure of all the ways that it may or may not be used after you've completed the activity/flow.

I opted to make the change in `useEntitiesSync`, and in doing so I noticed that the code had the potential to infinite-loop in one of the conditional branches. So I fixed that as well.

sultanofcardio (approved):
Lgtm

PR #600, opened 2025-03-21
fix: Prolific - public link deleted, user is unable to submit

hamzace:
### 📝 Description

🔗 [Jira Ticket M2-8921](https://mindlogger.atlassian.net/browse/M2-8921)

The issue when a user tries to submit a study after the Public Link has been removed while they are answering the survey. The expected behavior is that the system should display an error message, apologize for the inconvenience, and instruct the participant to enter "NOCODE" in the Prolific interface. However, in practice, the user is unable to submit the study when the Public Link is no longer available.

This PR fixes the issue by showing an error message if the submission fail.

The backend responds with 403 forbidden, since the applet is no longer available to public.


mbanting:
Finally was able to figure out how to test this and confirmed that backend responds with 403 but web app now handles with error banner to user with NOCODE message, as expected
image

src/widgets/Survey/ui/PassingScreen.tsx
```diff
@@ -193,9 +195,21 @@ const PassingScreen = (props: Props) => {
...
+    handleProlificError(() => {
+      if (error.response?.status === 400) {
+        addErrorBanner(t('prolific.alreadyAnswered'));
+      } else if (error.response?.status === 403) {
+        addErrorBanner(t('prolific.nocode'));
+      } else {
+        throw error;
+      }
+    });
   };
```
mbanting, line 213: With the growing amount of prolific-specific functionality, I think this would be a better fit in the `useProlific` hook. This ensures separation of concerns and modularity. The `PassingScreen` component is responsible for managing the survey flow, user interactions, and UI rendering. Adding Prolific-specific logic here introduces unrelated complexity and mixes concerns. By moving Prolific-specific logic to the `useProlific` hook, the component can focus solely on its primary responsibilities

My suggestion is to move this to the hook
```
const handleProlificSubmitError = useCallback(
  (error: AxiosError, addErrorBanner: (message: string) => void) => {
    if (!prolificParams) {
      return;
    }

    if (error.response?.status === 400) {
      addErrorBanner(t('prolific.alreadyAnswered'));
    } else if (error.response?.status === 403) {
      addErrorBanner(t('prolific.nocode'));
    } else {
      throw error;
    }
  },
  [prolificParams, t],
);
```
Then simplify the onSubmitError to this suggested code 
```suggestion
  const onSubmitError = (error?: AxiosError) => {
  closeSubmitModal();
  if (!error) {
    return;
  }

  handleProlificSubmitError(error, addErrorBanner);
};
```

mbanting (changes requested):

mbanting (approved):
@hamzace great work! Thanks!

PR #613, opened 2025-04-23
feat: Update RequestHealthRecordData to be skippable (M2-9059)

andrevitalb:
- [X] Tests for the changes have been added

### 📝 Description

🔗 [Jira Ticket M2-9059](https://mindlogger.atlassian.net/browse/M2-9059)

This PR adds updates to `RequestHealthRecordData` to be a skippable item.

A separate update to the current development flow was also included, adding the `lint-staged` package to the project, updating the current Husky pre-commit hook flow which consists of 4 steps:

1. Lint
2. Format
3. Test
4. Build

After this PR, the first 2 steps are handled by lint-staged, and the 2 last steps are removed from the hook and only handled within the repo's CI GitHub action workflow.

### 📸 Screenshots

https://github.com/user-attachments/assets/898d05fd-3323-4c16-870b-fb9a0c6df305

### 🪤 Peer Testing

To test this feature, an activity with an EHR item should exist and it should have the "Skippable Item" option turned on. In order to achieve this, three things must be true:

1. The [BE PR](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1824) should be merged. A workaround is to point the local BE to that PR's branch.
2. The [Admin PR](https://github.com/ChildMindInstitute/mindlogger-admin/pull/2070) should be merged. Similar to the last point, a workaround is to point the local admin to that PR's branch.
3. The `enableEHRHealthData` should be set to `"active"` in the admin panel, in order to enable the "Request Health Record Data" item type in the activity editor.

### DIFF
--- a/src/entities/activity/lib/types/item.ts
+++ b/src/entities/activity/lib/types/item.ts
@@ -460,7 +460,7 @@ export interface RequestHealthRecordDataItem extends ActivityItemBase {
 
 export type RequestHealthRecordDataItemConfig = {
   removeBackButton: boolean;
-  skippableItem: boolean;
+  skippableItem?: boolean;
 };
 
 export type RequestHealthRecordDataValues = {
--- a/src/entities/activity/lib/types/item.ts
+++ b/src/entities/activity/lib/types/item.ts
@@ -460,7 +460,7 @@ export interface RequestHealthRecordDataItem extends ActivityItemBase {
 
 export type RequestHealthRecordDataItemConfig = {
   removeBackButton: boolean;
-  skippableItem: never;
+  skippableItem: boolean;
 };
 
 export type RequestHealthRecordDataValues = {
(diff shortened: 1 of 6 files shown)


src/entities/activity/lib/types/item.ts
```diff
@@ -460,7 +460,7 @@ export interface RequestHealthRecordDataItem extends ActivityItemBase {
 
 export type RequestHealthRecordDataItemConfig = {
   removeBackButton: boolean;
-  skippableItem: never;
+  skippableItem: boolean;
```
farmerpaul, line 463: For backwards-compatibility with the previous configuration setup (it's not strictly needed, but just to align with my proposed BE fix):

```suggestion
  skippableItem?: boolean;
```

farmerpaul (approved):
Looks great, thanks for fixing husky! Also for adding those unit tests. 💪🏻

src/entities/activity/lib/types/item.ts
andrevitalb: Updating 🤓

PR #614, opened 2025-04-23
fix: Remove unwanted whitespace in component test id (M2-8988)

andrevitalb:
### 📝 Description

🔗 [Jira Ticket M2-8988](https://mindlogger.atlassian.net/browse/M2-8988)

This PR adds a simple fix to an issue that was causing extra whitespace to be rendered in some components' test IDs.

A simple `trim()` prevents extra whitespace from being rendered when the conditional test id assignment returns `undefined`.

### 🪤 Peer Testing

This was causing some tests to fail, so running the whole test suite successfully, without any issues related to whitespace or missmatched test ids, should work as a clear indicator this specific issue is solved.

### DIFF
--- a/src/shared/ui/Items/SelectBase/SelectBaseBox.tsx
+++ b/src/shared/ui/Items/SelectBase/SelectBaseBox.tsx
@@ -37,8 +37,6 @@ export const SelectBaseBox = (props: Props) => {
     ? hoverBackgroundColor
     : Theme.colors.light.neutural90;
 
-  const dataTestId = `select-box ${props.color ? `bgcolor-${props.color}` : ''}`.trim();
-
   return (
     <Box
       display="flex"
@@ -52,7 +50,7 @@ export const SelectBaseBox = (props: Props) => {
       border={`2px solid ${borderColor}`}
       bgcolor={props.color ? props.color : backgroundColor}
       onClick={props.onHandleChange}
-      data-testid={dataTestId}
+      data-testid="select-box"
       sx={{
         ...(props.sx ?? {}),
         transition: 'background-color 0.2s ease-in-out',
--- a/src/shared/ui/Items/SelectBase/SelectBaseBox.tsx
+++ b/src/shared/ui/Items/SelectBase/SelectBaseBox.tsx
@@ -37,6 +37,8 @@ export const SelectBaseBox = (props: Props) => {
     ? hoverBackgroundColor
     : Theme.colors.light.neutural90;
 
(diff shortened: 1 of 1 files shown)


src/shared/ui/Items/SelectBase/SelectBaseBox.tsx
```diff
@@ -37,6 +37,8 @@ export const SelectBaseBox = (props: Props) => {
     ? hoverBackgroundColor
     : Theme.colors.light.neutural90;
 
+  const dataTestId = `select-box ${props.color ? `bgcolor-${props.color}` : ''}`.trim();
```
farmerpaul, line 40: I'm not certain that test IDs should ever contain _any_ whitespace, but I may be wrong. While your fix will eliminate trailing whitespace as requested, E2E tests could possibly fail again when the **Set Color Palette** option is enabled in the admin, because doing so changes the test IDs:

CleanShot 2025-04-23 at 08 03 37@2x

When this setting is enabled, the options end up having test IDs like `select-box bgcolor-#ffadad`:

CleanShot 2025-04-23 at 08 09 48@2x

I suspect it may not be desirable for the test IDs to change based on its palette setting, but I could very well be wrong. @egodoy-metalab what do you think?

farmerpaul (commented):
I'm a bit uncertain about the requirements as articulated in the Jira task. I've raised one concern in a comment below (and it could be a non-issue, pending @egodoy-metalab's feedback). But I'm also unclear about where this issue is occurring.

The Jira task says:

> This issue is happening in: 
> - Textbox
> - Slider
> - Radio button

But the `SelectBaseBox` component, where @andrevitalb made a change in this PR, only affects these item types:

- Single Selection (radio buttons)
- Multiple Selection (checkboxes)

This change would have no impact on textboxes or sliders, but both of those item types are mentioned in the Jira issue description. Without knowing exactly what test IDs relating to textboxes and sliders contain whitespace, we don't know what code needs to be fixed.

@egodoy-metalab I see that while Katy is the person who created the Jira ticket, but I think it was based on information you provided. Could you have a look at [M2-8988](https://mindlogger.atlassian.net/browse/M2-8988) and ensure that it's articulated accurately? Thank you!

src/shared/ui/Items/SelectBase/SelectBaseBox.tsx
andrevitalb: After syncing with @egodoy-metalab, we're getting rid of the dynamic test ID for this case

PR #624, opened 2025-05-17
feat: EHR error handling (M2-8955)

andrevitalb:
- [ ] Tests for the changes have been added

### 📝 Description

🔗 [Jira Ticket M2-8955](https://mindlogger.atlassian.net/browse/M2-8955)

This PR adds error handling for EHR (1UpHealth-related) errors, matching the changes added in [this BE PR](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1850).

An update to 1UpHealth's token fetching endpoint was also added to match latest BE changes. The API route switched from `/integrations/oneup_health/applet/${appletId}/submissions/${submitId}/token` -> `/integrations/oneup_health//applet/{applet_id}/submission/{submit_id}/activity/{activity_id}/token`.

An additional endpoint was created for the BE to refresh 1UpHealth tokens.


farmerpaul (commented):
Hey @andrevitalb, nice work so far! I haven't done any peer testing and I'll wait till you fill out that section before I do (since maybe you have some tricks up your sleeve there). Just had one nit, and I was wondering if you could also clean up the commit history for this branch when you have a chance? Seeing a lot of my own commits in it.

src/entities/activity/api/useOneUpHealthTokenQuery.ts
```diff
@@ -4,18 +4,19 @@ type FetchFn = typeof OneUpHealthService.retrieveTokenBySubmitId;
...
     () =>
       OneUpHealthService.retrieveTokenBySubmitId({
         appletId,
         submitId: submitId as string,
+        activityId: activityId as string,
       }),
     {
-      retry: 4,
+      retry: false,
       enabled: !!submitId,
```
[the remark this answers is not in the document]
andrevitalb: Makes sense! Updating 🤓

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
```diff
@@ -29,14 +52,176 @@ export const OneUpHealthStep: FC = () => {
...
+          err?.status === 502 &&
+          ((err?.evaluatedMessage || (err.response.data.result?.[0]?.message ?? '')).includes(
+            'token has expired',
+          ) ??
+            false),
+        title: t('oneUpHealth.tokenExpired.title'),
+        message: t('oneUpHealth.tokenExpired.message'),
+        banner: t('oneUpHealth.tokenExpired.banner'),
+        logMessage: 'Token expired error',
+      },
```
farmerpaul, line 140: This is just my opinion, but hear me out: I don't think that the user should be alerted about an expired authentication token if it's automatically refreshed anyway. When the token expires, the alert message currently shown is so brief that it's actually impossible to read either the alert or the error message below it before the iframe is automatically reloaded with the refreshed token.

To my point: we don't alert the user about expired MindLogger app auth token, either in the web app or the admin app. We just have integrated token refresh functionality that provides a seamless user experience. I don't see why we should treat this situation any differently, but maybe I'm missing something.

Thoughts @mbanting?

farmerpaul (commented):
Great work, thanks for the handy peer testing tool! I added a comment about token refresh that I'd be interested in hearing others' opinions on.

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
```diff
@@ -63,35 +248,58 @@ export const OneUpHealthStep: FC = () => {
...
           <Loader />
         </Box>
       )}
+      {errorType && (
+        <Box
+          display="flex"
+          flexDirection="column"
+          alignItems="center"
+          justifyContent="center"
+          height="400px"
```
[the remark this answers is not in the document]
andrevitalb: Thanks! Updating

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
mbanting: @andrevitalb and I had a conversation yesterday and had the same thought. 
> Hmm, ideally the whole error would be transparent from the user (ie. the app would detect it and without showing the error, attempt to resolve itself by refreshing the token). 

Is there a need to show this error message? 
We aren't waiting for any user action to refresh the token, are we?
What is the UX like if we simply don't show the error message while the app is refreshing the token behind the scenes?

andrevitalb: Since this is happenning automatically, we aren't waiting for anything. Initial addition of the error visibility was following the ACs. Updated to just keep the error logging for visibility, removing the error reflecting on any UI changes, other than refreshing the iframe once token has been refreshed

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
```diff
@@ -29,14 +52,177 @@ export const OneUpHealthStep: FC = () => {
...
+  }, [refreshTokenValue, refreshToken]);
 
   // Instantiate channel and iframeRef variable
   const messageChannelRef = useRef<MessageChannel>();
   const iframeRef = useRef<HTMLIFrameElement>(null);
   const [isIframeLoaded, setIsIframeLoaded] = useState(false);
 
+  const errorTypes: Record<OneUpHealthErrorType, OneUpHealthErrorConfig> = useMemo(
+    () => ({
+      geographicRestriction: {
```
sultanofcardio, line 117: I don't think this error message will ever happen. This response is coming from the MindLogger API, which presumably will be located on a server within the US. It may become an issue if we ever decide to deploy to servers outside the US though.

However, currently when I try this in the Amplify environment with no VPN, the token fetch from the BE succeeds and I'm presented with an iframe with a forbidden message. Do you also get this?

https://github.com/user-attachments/assets/e73bc892-ca76-4f68-ab02-4428587656bb

farmerpaul: Ah, really good point @sultanofcardio. Practically speaking, _this ☝🏻_  is how the geographic restriction will come into play in the web app. It's when the client is located outside the US that it fails, with this (presumably) difficult-to-detect error. The token will come back fine from the BE, but the iframe will still refuse to load.

I guess it returns JSON instead of HTML in this case. I wonder if theres a way for the web app containing the iframe to discover the failure…

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
```diff
@@ -29,14 +52,177 @@ export const OneUpHealthStep: FC = () => {
   const submitId = groupProgress?.submitId;
 
   // Fetch the token using our custom hook
-  const { data, isLoading } = useOneUpHealthTokenQuery({ appletId, submitId });
-  const accessToken = data?.data?.result?.accessToken;
+  const { data, isLoading, error } = useOneUpHealthTokenQuery(
+    { appletId, submitId, activityId },
+    { refetchOnWindowFocus: false },
```
sultanofcardio, line 57: We should set a stale time here, otherwise we throw away the cached value when the query refetches (and fails) on component re-mount. You probably know a more appropriate value, I just put `60000` as a placeholder
```suggestion
    { refetchOnWindowFocus: false, staleTime: 60000 },
```

https://github.com/user-attachments/assets/cd117146-e150-4786-9e12-424af4e8c84a

---

The query refetch fails because the BE tries to recreate the 1Up user (applet + activity + submit ID), but they already exist so it fails. Not sure why it does this

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
sultanofcardio: With all the security controls around cross-origin iframes, I'm not so sure this is even possible unless the iframe sends a message through the channel when this happens

sultanofcardio:
@andrevitalb I suggest putting the debug panel back into the commit history so QA can use it for testing in amplify. We can always remove it before merging.

The detached head method also prevents us from getting any subsequent changes

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
farmerpaul: Perhaps an XHR request to the same URL could be made ahead of time, to determine if it returns a 403 status code. If it does, trigger this error message, else display the iframe?

mbanting: ☝️ that was my initial thinking. Almost like a ping.

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
mbanting: May have been a little confusing, but AC for banner didn't include **token expiration**:

> Users will see an error banner (banner and wording TBC by product/design) when failures caused by the 1UpHealth / MindLogger system integration occurs. This includes:
> 
> - 1UpHealth availability and access token generation errors. 
> 
> - MindLogger / 1UpHealth integration endpoints (eg. /integrations/oneup_health//applet/{applet_id}/submission/{submit_id}/activity/{activity_id}/token)
> 
> - iframe “forbidden” error when the client browser is detected to be running outside of the USA

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
sultanofcardio: I thought the same, but CORS will block this request

mbanting:
> @andrevitalb I suggest putting the debug panel back so QA can use it for testing in amplify. We can always remove it before merging.
> 
> The detached head method also prevents us from getting any subsequent changes

Good idea. I also suggested the option of 

> @andrevitalb I suggest putting the debug panel back so QA can use it for testing in amplify. We can always remove it before merging.
> 
> The detached head method also prevents us from getting any subsequent changes

This is an option. Alternatively, QA Notes were recently added to the ticket that suggests how QA can test without this debug panel.

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
andrevitalb: Indeed. This was an approach I had attempted to try finding a workaround without much success. Right now, I'm following an approach that checks for the iframe's content after it's attempted to load. One caveat is that this approach only allows for unsuccessfull loads, without distinguishing the actual reason behind it, so we can't actually know if the issue was a geographic restriction or any other for that matter

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
andrevitalb: Indeed. I can go ahead and a couple changes on this to improve this functionality. What do you think @sultanofcardio @farmerpaul?

sultanofcardio: @andrevitalb Yes, I think that would be good

src/entities/activity/ui/items/RequestHealthRecordDataItem/OneUpHealthStep.tsx
farmerpaul: @andrevitalb I just had a thought. In the mobile app, we do an API call to `https://system-search.1up.health/api/search` passing the access token as a header in order to populate the system search results. If the user is outside of the USA, then I think this returns a forbidden error state. So we could use this as an early check to determine if the user won’t be able to load the iframe, and if not, then display the user-friendly “you’re not in the USA” error state. What do you think?

You'd need to add the `ONEUP_HEALTH_SYSTEM_SEARCH_API_URL` env var to the web app, and document it in the README, etc. You should be able to copy-paste that and adapt the querying code from the mobile app.

andrevitalb: @farmerpaul I'll try this approach. Will add results here.

andrevitalb: At this point, I believe the best way to solve the iframe issue is to not use it at all. Changing the implementation to do the system search in the BE and return results, similar to what we do for the app, just with the BE in the middle.

mbanting: Great discussion and thank you all for thinking of different ways to solve this. 

Eventually we can change out the iframe but not for this MVP release. 
As mentioned previously, the iframe was chosen as part of the leanest path to an E2E test, one that would allow us to deliver by the target date. 

Due to the additional complexities for this specific AC, we've since separated this to a separate ticket (M2-9330) so as to not block the rest of the related but separate and independent AC from moving forward. I'll make a note on that ticket to reach out to 1up to see if there are any endpoints that are available that
- are not subject to CORS
- but are subject to US geographical access

@andrevitalb if it hasn't been done already, could you remove any implementation of this AC from the PR. That way this AC is no longer outstanding and we can move one step closer to merging this PR?

Thank you!

andrevitalb: @mbanting Updated. I just kept the initially added implementation that receives the error from the BE and handles it. The corresponding error screen is also present, but won't be triggered by the iframe errors

farmerpaul (approved):
Nice keyboard shortcut addition. Great work!

PR #629, opened 2025-05-28
feat: Brand intro banner (M2-9068)

andrevitalb:
- [X] Tests for the changes have been added

### 📝 Description

🔗 [Jira Ticket M2-9068](https://mindlogger.atlassian.net/browse/M2-9068)

This PR introduces a brand new Rebrand banner to the application as a first notice for the upcoming branding changes.

### 🪤 Peer Testing

Testing is pretty simple.

The banner should be visible on these screens:
- All unauthenticated routes:
- Authenticated routes, excluding:
  - Any screen related to an active assessment, including the submission screens

The dismissed state is tracked separately for auth & non-auth routes.

Dismissing the banner in a non-auth route should keep the state persistent throughout the use of these routes, without login/logout sessions disrupting it.

On the other hand, dismissing in an auth route should keep the state persistent for the logged in user throughout login sessions.


src/shared/ui/Banners/RebrandBanner/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+const NON_AUTH_DISPLAY_ROUTES = [
+  /^\/login$/,
+  /^\/signup$/,
+  /^\/forgotpassword$/,
+  /^\/password-recovery$/,
+];
+const AUTH_DISPLAY_ROUTES = [
+  /^\/protected\/(profile|settings)$/,
+  /^\/protected\/applets(\/[^/]+)?$/,
+];
```
farmerpaul, line 30: I clarified that AC, which had been copy-pasted from the EHR admin banners task and didn't actually totally make sense originally 🤦🏻‍♂️. I updated it to be all screens except when performing an assessment, same as the mobile app. This code may already happen to be in alignment, but not sure if you want to update it to be more about excluding those screens, rather than including all these screens? I did [this](https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/977/files#diff-a276ac6d707c3f41958f37e1f7551bd95eb4a91f9fbdca73077a43111ac5ca72R3-R7) in the mobile app.

src/shared/ui/Banners/RebrandBanner/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+      setIsRebrandBannerActive(!globalDismissed);
+      return;
+    }
+
+    // For logged-in users
+    if (userId) {
+      const userDismissed = localStorage.getItem(getDismissedKey(userId));
+      setIsRebrandBannerActive(!userDismissed);
+    }
+  }, [isAuthorized, userId]);
```
farmerpaul, line 60: Nit: a little simpler?

```suggestion
  const dismissedKey = isAuthorized && userId ? getDismissedKey(userId) : GLOBAL_DISMISSED_KEY;

  useEffect(() => {
    const dismissed = localStorage.getItem(dismissedKey);
    setIsRebrandBannerActive(!dismissed);
  }, [dismissedKey]);
```

src/shared/ui/Banners/RebrandBanner/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+    : NON_AUTH_DISPLAY_ROUTES.some((route) => route.test(location.pathname));
+
+  const handleDismiss = () => {
+    // For auth screen or when user is not logged in yet
+    if (!isAuthorized) {
+      localStorage.setItem(GLOBAL_DISMISSED_KEY, 'true');
+    } else if (userId) {
+      // Set user-level dismissal
+      localStorage.setItem(getDismissedKey(userId), 'true');
+    }
```
farmerpaul, line 73: If you do the previous simplification, then `dismissedKey` is already available:
```suggestion
    localStorage.setItem(dismissedKey, 'true');
```

src/shared/utils/hooks/useRebrandBanner/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export const useRebrandBanner = () => {
+  const { addBanner } = useBanners();
+
+  useEffect(() => {
+    // Use a timeout to ensure this runs after any other banners are added
+    // This will make the RebrandBanner appear on top
+    setTimeout(() => {
+      addBanner('RebrandBanner', {});
+    }, 0);
```
farmerpaul, line 17: I'm not quite sure how the timeout is supposed to consistently achieve the intended ordering. I was actually able to do this in one instance:

CleanShot 2025-05-28 at 19 26 29@2x

This seems to happen when you perform these steps:
1. Complete an assessment.
2. When you're back on the single applet screen, the success banner appears, but curiously, the rebrand banner _does not_.
3. Reload the page. Both banners appear, with the rebrand banner at the bottom.

farmerpaul (approved):
Nice job @andrevitalb! A couple nits, and a couple minor behaviour bugs I described in a comment that I don't consider bad enough to be a blocker. Feel free to perfect the behaviour if you like, but I'm happy to preapprove!

src/shared/ui/Banners/RebrandBanner/index.tsx
andrevitalb: Ooh, I like this. Updating it 🤓

src/shared/ui/Banners/RebrandBanner/index.tsx
andrevitalb: Oh, inverting the logic here. Makes sense. Updating

src/shared/utils/hooks/useRebrandBanner/index.ts
andrevitalb: Yep, a tad inconsistent. Updating this.

After the changes, the small timeout (10ms) helps ensure proper ordering by slightly delaying the addition of the rebrand banner, allowing it to be added after other banners like the success banner.

Also, to fix the issue where the banner was only showing up after reloading, made the hook responsive to navigation events.

farmerpaul: I'm a little unclear about Yanina's [latest comment in Jira](https://mindlogger.atlassian.net/browse/M2-9068?focusedCommentId=33066). She seemed to state the the rebrand banner is showing at the top, but then provided screenshots of it showing at the bottom.

Anyway I'm assuming your changes fixed it, but if it's indeed still happening, you could consider adding explicit banner ordering (like in the [mobile app](https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/977/commits/8ab9f874db29e88a9f4d6b150b66ea9a2a09e8c9)).

farmerpaul: Oh, I re-read it – never mind, she's saying that _other banners_ appear at the top, above the rebrand banner. Which I guess is what your latest changes ensure is happening consistently.

I actually interpreted the designs differently in my latest changes to the mobile app, placing the rebrand banner consistently at the top, while other temporary notifications appear below it, which aligns more closely with the admin app. What do you think?

andrevitalb: Yep, I had added an approach that relied on timing. Liking yours better. Updating here 🤓

PR #640, opened 2025-06-19
feat: Add EHR redirect interstitial page for desktop and mobile (M2-9361)

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-9361](https://mindlogger.atlassian.net/browse/M2-9361)

This PR adds an interstitial page at the path `/ehr-complete` to act as the redirect path for 1UpHealth apps. Users who start an EHR in the mobile app on Android will eventually get redirected here and will be required to click a button to trigger the re-opening of the app.

The motivation for this is that Android apps require an explicit user interaction in order to intercept a deep link. 1UpHealth uses an automatic client side redirect that lacks user interaction, so the Android app fails to open.

Here's a demonstration of this functionality on Android:

https://github.com/user-attachments/assets/09f2cb0e-52a5-4bc0-81dc-e7b1684531d4

> [!NOTE]
> There's a bug in this video where the background image was not displaying at the time the video was recorded. This has since been fixed with 17183e9ca38aff7b3f2687a9ac6c7996e7d0b174

#### iOS support

Given that 1UpHealth apps only allow one redirect URL, both iOS and Android users will be sent to this interstitial screen at `/ehr-complete`. The current redirect URL (`/active-assessment`) was set up as a deep link in the app, but will no longer be activated due to this change. That left us with two options:
1. Allow iOS users to go to the interstitial screen in the browser and press the button, just like Android users
2. Add `/ehr-complete` as a deep link path in the iOS app

I initially went with option 1, but quickly ran into a roadblock preventing it from working: iOS does not support websites launching deep links that go to the same website - the browser retains controls and the app does not open. I therefore added support for `/ehr-complete` as a deep link path on iOS by updating the `apple-app-site-association` file in this repo.

image

Deep link support for that path is completed in https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/998, and therefore that PR is required to fully test this change. What this means is that iOS users will not see this page, as shown in the following video:

https://github.com/user-attachments/assets/a42d10a2-7c49-4b6a-802f-03a900523887

#### Browser support

Similar to iOS users, any users who initiate an EHR from a web browser will eventually land on this page. This necessitated a deviation from the design in three (3) key ways:

1. The addition of a "continue in browser" button - This button will not launch the mobile app for Android users who started their EHR in the web browser, even if they have the app installed
2. The rewording of the modal copy to accommodate the use case of continuing in the browser
3. The modification of the background image to support wider viewports

Here are some examples of flows taking place in web browsers.

##### Desktop

https://github.com/user-attachments/assets/84f71cb5-630d-4242-894a-410de8ab5ceb

> [!NOTE]
> Technically both buttons will function to "continue in browser" here, as the mobile app is not installed. This is demonstrated at the end of the video

##### Mobile (Android)

https://github.com/user-attachments/assets/a3ced7ee-6fc2-41f9-b019-7a47ced629c8

> [!NOTE]
> Observer at the end of the video the demonstration that the "open in app" button correctly functions to open the mobile app

### 🪤 Peer Testing

> [!WARNING]
> If you are keen on testing this thoroughly, note that it is quite involved

#### iOS test

1. Create a 1UpHealth app that uses the following URL as its redirect URL: https://pr-640.d15zn9do8xbzga.amplifyapp.com/ehr-complete
2. Edit your local API's `.env` file to use the `ONEUP_HEALTH__CLIENT_ID` and `ONEUP_HEALTH__CLIENT_SECRET` values
3. Create an applet with an activity containing an EHR item
4. Check out https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/998, and make the following changes in that repo
5. Edit the `ios/MindloggerMobileDevDebug.entitlements` file to add two entries for the `pr-640.d15zn9do8xbzga.amplifyapp.com` domain
    image
6. Edit your `.env.dev` file to set the value of the `ONEUP_HEALTH_CLIENT_ID` to the same as the API's
7. Build and launch the iOS app
8. Complete the EHR activity and confirm that the `/ehr-complete` URL does not open the web page - it should go straight to the app allowing you to complete the activity

#### Android test

1. Create a 1UpHealth app that uses the following URL as its redirect URL: `http[s]://<your_lan_ip>:5173/ehr-complete`
2. Edit your local API's `.env` file to use the `ONEUP_HEALTH__CLIENT_ID` and `ONEUP_HEALTH__CLIENT_SECRET` values
3. Check out https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/998, and make the following changes in that repo
4. Edit `android/app/build.gradle` to change the dev secondary domain from `web-dev.cmiml.net` to `<your_lan_ip>`
    image
5. Edit your `.env.dev` file to set the value of `API_URL` to `http://<your_lan_ip>:8000`. Also set the value of the `ONEUP_HEALTH_CLIENT_ID` to the same as the API's
6. Build and launch the Android app
7. Go into the app info and enable your IP address as a deep link domain (as shown in the video above)
8. Launch the app and initiate the activity. Confirm that the `/ehr-complete` URL goes to the expected webpage
9. Click the "Open app" button. Confirm that the app opens, and you are able to complete and submit the activity

#### Web browser tests

1. On the same android device as above, open the web browser and go to your local instance of the web app
2. Log in and complete the EHR activity in the web browser
3. When you get to the interstitial screen, click "continue in browser". Confirm that the mobile app is not opened
4. Confirm that you are able to complete and submit the activity
5. Perform the same test in your regular desktop browser and confirm that you are able to complete and submit the activity

### ✏️ Notes

There is a version bump of the `i18next-browser-languagedetector` package in the PR, which is a result of me trying to figure out why the language detection doesn't work. As it turns out, it does work. It's just not straightforward. In order to get the Curious web app to auto-select the user's language, the language has to be set as the topmost preferred language in their web browser

image

> [!NOTE]
> Screenshot from Chrome language settings (`chrome://settings/languages`)

Then, they would have to be opening the web app for the first time, so that there would not be an existing language setting preference in local storage. You can simulate this by doing the browser language change, clearing the site data, and reloading the page.

After discovering this, I decided to leave in the change, as it seems to still work 🤷


package.json
```diff
@@ -12,7 +12,6 @@
 		"lint:fix": "eslint src --fix",
 		"prettier:check": "prettier src --check",
 		"prettier:fix": "prettier src --write",
-		"prepare": "husky install",
```
andrevitalb, line 15: Why are we removing this? We're still using Husky in the project

andrevitalb (approved):
Code looks good. Moving to peer testing

package.json
andrevitalb: Husky runs a hook on every commit 👀. See [pre-commit](https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/dev/.husky/pre-commit). Or, am I missing something here?

farmerpaul: Yeah, the script should be deleted too, if Husky is removed (which I'm in favour of). Kenroy and I just don't find the pre-commit helpful at all, so we turn off cause it needlessly slows down development.

mbanting:
I'm a bit lost on how this is working. 1up is currently configured to redirect to MindLogger/Curious via `/active-assessment`
How will iOS and Android users be sent to this interstitial screen at `/ehr-complete`?

sultanofcardio:
> @sultanofcardio I'm a bit lost on how this is working. 1up is currently configured to redirect back to MindLogger/Curious via `/active-assessment` How will iOS and Android users be sent to this interstitial screen at `/ehr-complete`?

@mbanting after this is merged, we'll need to change the redirect path in the 1UpHealth app to point to this interstitial screen at `/ehr-complete`. In these videos, I create my own 1UpHealth app that redirects there

PR #689, opened 2025-12-22
feat(M2-10000): implement MFA login flow

maxto024:
# PR Description - MFA Implementation

## Summary

Implements Multi-Factor Authentication (MFA) for enhanced login security in MindLogger Web. Users with MFA enabled must provide a 6-digit TOTP code or recovery code after entering their credentials.

## Jira Ticket
[M2-10000](https://mindlogger.atlassian.net/browse/M2-10000)

## Screenshot

Screenshot 2025-12-22 at 5 52 38 PM
Screenshot 2025-12-22 at 5 52 46 PM
Screenshot 2025-12-22 at 5 52 54 PM
Screenshot 2025-12-22 at 5 53 02 PM

### Core Features
- **Route-based MFA Flow**: Separate pages for TOTP and recovery code verification
- **MFAContext Provider**: Global React Context for secure session state management
- **TOTP Verification**: 6-digit code verification with auto-submit functionality
- **Recovery Codes**: Backup authentication method with proper formatting (XXXXX-XXXXX)
- **Backend-driven Session Expiry**: No local timers, API determines session validity
- **Rate Limiting**: Maximum 5 attempts with warning after 3 attempts
- **Route Guards**: MFA pages redirect to login if no active session

### Technical Implementation
- **State Management**: MFAContext with useReducer (in-memory only - security best practice)
- **Custom Hook Pattern**: `useMFAVerification` provides shared logic for both forms
- **API Integration**: Endpoints for TOTP and recovery code verification
- **Form Validation**: Zod schemas with real-time validation
- **Auto-submit**: Automatic form submission when 6 digits entered

### UI Implementation (Figma-compliant)
- Card container: 473px width, 32px padding, 16px border-radius
- Surface background with surfaceVariant border
- Proper typography variants and spacing

---

## Routes

| Route | Component | Purpose |
|-------|-----------|---------|
| `/login` | LoginPage | Email/password entry |
| `/auth/verify-mfa` | MFAVerifyPage | TOTP 6-digit verification |
| `/auth/verify-recovery` | MFARecoveryPage | Recovery code verification |

---

## Files Created

| File | Purpose |
|------|---------|
| `src/features/Login/lib/MFAContext.tsx` | Global context provider for MFA session state |
| `src/features/Login/model/mfa.types.ts` | MFA TypeScript type definitions |
| `src/features/Login/model/mfa.schema.ts` | Zod validation schemas |
| `src/features/Login/lib/useMFAVerification.ts` | Core verification logic hook |
| `src/features/Login/ui/MFAForm.tsx` | TOTP 6-digit form with auto-submit |
| `src/features/Login/ui/RecoveryCodeForm.tsx` | Recovery code form with auto-format |
| `src/pages/MFAVerify/index.tsx` | TOTP verification page |
| `src/pages/MFARecovery/index.tsx` | Recovery code page |

## Files Modified

| File | Changes |
|------|---------|
| `src/app/providers/index.tsx` | Added MFAProvider wrapper |
| `src/pages/UnauthorizedRoutes.tsx` | Added MFA page routes |
| `src/shared/constants/routes.ts` | Added verifyMFA and verifyRecovery routes |
| `src/shared/api/types/authorization.ts` | Added MFA API types |
| `src/shared/api/services/authorization.service.ts` | Added MFA verification endpoints |
| `src/i18n/en/translation.json` | Added MFA translations |
| `src/features/Login/ui/LoginForm.tsx` | Added onMFARequired callback |
| `src/pages/Login/index.tsx` | Handle MFA redirect via context |

---
## Security Considerations

- ✅ MFA tokens stored in React Context (in-memory only)
- ✅ NOT persisted to Redux or localStorage
- ✅ Backend-driven session expiry
- ✅ Route guards redirect to `/login` if no MFA session
- ✅ loginPassword retained in-memory for encryption key derivation
- ✅ No sensitive data logged to console

---

## How to Test

1. Login with MFA-enabled user
2. Verify navigation to `/auth/verify-mfa`
3. Test invalid code → inline error (no redirect)
4. Test "I can't access my authenticator app" → navigates to `/auth/verify-recovery`
5. Test "Back" link on recovery page → returns to TOTP form
6. Test session expiry → shows expiry error with "Back to login"
7. Test successful verification → navigates to dashboard
8. Verify card styling matches Figma specs

---

## Backwards Compatibility

- ✅ Non-breaking change
- ✅ Users without MFA continue to login normally
- ✅ Existing auth tokens remain valid

---

## Documentation

- [docs/MFA_IMPLEMENTATION.md](./MFA_IMPLEMENTATION.md) - Complete implementation details

---

## Checklist

- [x] Route-based MFA navigation
- [x] MFAContext for secure state management
- [x] TOTP verification with auto-submit
- [x] Recovery code verification
- [x] Backend-driven session expiry
- [x] Route guards for MFA pages
- [x] Figma-compliant UI styling
- [x] Translations aligned with Figma
- [x] TypeScript types properly defined
- [x] Zod validation schemas
- [x] Documentation updated

[M2-10000]: https://mindlogger.atlassian.net/browse/M2-10000


src/app/providers/index.tsx
```diff
@@ -16,11 +18,13 @@ function Providers({ children }: Props) {
       <CssBaseline />
       <MUIThemeProvider>
         <RouteProvider>
-          <ReduxProvider>
-            <ReactQuery>
-              <DateLocalizationProvider>{children}</DateLocalizationProvider>
-            </ReactQuery>
-          </ReduxProvider>
+          <MFAProvider>
```
divbzero: Is there a way we could avoid implementing MFA using a global `MFAProvider`? The MFA auth flow occurs over a few screens, like the existing password auth flow, so I think we should be able to implement without wrapping the whole app in a `MFAProvider`.

src/features/Login/ui/LoginForm.tsx
```diff
@@ -29,7 +32,18 @@ export const LoginForm = ({ locationState }: LoginFormProps) => {
 
   const { mutate: login, isLoading } = useLoginMutation({
     onSuccess(data, variables) {
-      const { user, token } = data.data.result;
+      const result = data.data.result;
+
+      // Check if MFA is required
+      if (isMFARequiredResponse(result)) {
+        if (onMFARequired) {
```
divbzero: We shouldn’t need an `if (onMFARequired)` check here. We should at least make `onMFARequired` a required prop and remove this check, but a better approach might be to move the `handleMFARequired` logic from **src/pages/Login/index.tsx** to here and avoid having to use an `onMFARequired` callback altogether.

src/pages/Login/index.tsx
```diff
@@ -17,11 +19,23 @@ function LoginPage() {
...
   const onCreateAccountClick = () => {
     Mixpanel.track({ action: MixpanelEventType.LoginScreenCreateAccountBtnClick });
   };
 
+  // Handle MFA required: store session and navigate to MFA page
+  const handleMFARequired = useCallback(
+    (mfaData: MFARequiredResponse, password: string) => {
+      setSession(mfaData.mfaToken, mfaData.mfaSessionId, password);
+      navigate(ROUTES.verifyMFA.path, {
+        state: { backRedirectPath: location.state?.backRedirectPath },
```
divbzero, line 60: Instead of `setSession` followed by `navigate`, we could consider doing the following:

1. Derive and store the private key using `password` first before we `navigate`. We don’t need to hold onto `password` any longer than necessary.
2. Pass `mfaToken` and `mfaSessionId` through navigation state.
    ```ts
    navigate(ROUTES.verifyMFA.path, {
      backRedirectPath: location.state?.backRedirectPath,
      mfaToken: mfaData.mfaToken,
      mfaSessionId: mfaData.mfaSessionId,
    });
    ```

With this approach, I think we could avoid using the global `MFAProvider` of `MFAContext`.

src/features/Login/lib/useMFAVerification.ts
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Helper to detect session expiry from backend error message
+ * Matches admin pattern: Auth.reducer.ts isSessionExpiredError()
+ */
+const isSessionExpiredError = (message: string): boolean => {
+  const lower = message.toLowerCase();
+  return (
+    lower.includes('session not found') ||
+    lower.includes('expired') ||
+    lower.includes('session expired')
```
divbzero: Are there exact error constants from **mindlogger-backend-refactor** that we could use instead of string matching error messages?

How `AuthErrorCode.MFA_SESSION_NOT_FOUND`?
https://github.com/ChildMindInstitute/mindlogger-backend-refactor/blob/30c5e9ed39fc3ef2565a46eaec9f0a4132bbc2ed/src/apps/authentication/constants.py#L13

Is that an error constant we could match against in `MFASessionNotFoundError`?
https://github.com/ChildMindInstitute/mindlogger-backend-refactor/blob/30c5e9ed39fc3ef2565a46eaec9f0a4132bbc2ed/src/apps/authentication/errors.py#L62-L65

(`MFATokenInvalidError` and `MFATokenExpiredError` might be other errors to match against. I have dug in to verify which error is relevant here.)

src/features/Login/lib/useMFAVerification.ts
```diff
@@ -0,0 +1,188 @@
...
+    if (isSessionExpired || displayError === 'mfaSessionExpired') {
+      return; // Don't clear terminal errors
+    }
+    setDisplayError(null);
+  }, [isSessionExpired, displayError]);
+
+  // Cleanup on unmount
+  const cleanup = useCallback(() => {
+    hasSessionExpiredRef.current = false;
+  }, []);
```
divbzero: It doesn’t look like we’re using this `cleanup` callback right now. Does it need to be called in **MFAForm.ts**? Or perhaps the cleanup isn’t necessary?

src/features/Login/ui/MFAForm.tsx
```diff
@@ -0,0 +1,235 @@
...
+          <TOTPInput
+            control={control}
+            onChange={handleCodeChange}
+            disabled={isSessionExpired}
+            error={!!helperMessage}
+            helperText={helperMessage || undefined}
+          />
+
+          <Box display="flex" justifyContent="center">
+            <button
```
divbzero: We should probably use `BaseButton` instead of `button` like we do for the `type="submit"` button below.

src/features/Login/ui/MFAForm.tsx
```diff
@@ -0,0 +1,235 @@
...
+          <BaseButton
+            type="submit"
+            variant="contained"
+            isLoading={isSubmitting}
+            disabled={isSessionExpired}
+            text={t('continue')}
+          />
+
+          <Box display="flex" justifyContent="center">
+            <button
```
divbzero: `s/button/BaseButton/` like above.

src/pages/MFARecovery/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+import { userModel } from '~/entities/user';
+import { RecoveryCodeForm, useLoginTranslation } from '~/features/Login';
+import { useMFAContext } from '~/features/Login/lib/MFAContext';
+import { ROUTES } from '~/shared/constants';
+import { variables } from '~/shared/constants/theme/variables';
+import { Box } from '~/shared/ui';
+
+function MFARecoveryPage() {
+  const { t } = useLoginTranslation();
```
divbzero: Do we need to add translations? Or will that be left for a future PR?

src/features/Login/lib/useMFAVerification.ts
```diff
@@ -0,0 +1,188 @@
...
+            setIsSessionExpired(true);
+            setDisplayError('mfaSessionExpired');
+          }
+          // Do NOT increment attempts for expired session
+          return false;
+        }
+
+        // Normal invalid code flow - increment attempts
+        onIncrementAttempts();
+        const newAttempts = attempts + 1;
```
divbzero: Could we use the attempts remaining returned by API responses instead of tracking the number of attempts on the client?

src/features/Login/lib/useMFAVerification.ts
```diff
@@ -0,0 +1,188 @@
...
+      if (!session) return false;
+
+      setIsSubmitting(true);
+      setDisplayError(null);
+
+      try {
+        const response =
+          type === 'totp'
+            ? await verifyTOTP({ mfaToken: session.token, totpCode: code })
+            : await verifyRecovery({ mfaToken: session.token, code });
```
divbzero: We might want to pass in `deviceId` as well?

divbzero (changes requested):
Generally looks good!

My biggest question is whether we could refactor the flow in the login form to:

1. Avoid holding onto `password` any longer than necessary.
2. Pass `mfaToken` and `mfaSessionId` to the MFA verification screen using navigation state, instead of using `MFAContext`.

This should allow us to simplify things. We could avoid introducing a `MFAContext` and a global `MFAProvider` that wraps the entire app.

src/app/providers/index.tsx
maxto024: We were having some render, and we needed a good way to pass the password without exposing to navigation that may put in the url i commented the `MFAProvidor`  also and index.ts why we may need to keep for now. 
please check the comment.

src/features/Login/lib/useMFAVerification.ts
maxto024: Removed unused cleanup, added TODO comments for backend error codes, deviceId, attempts

src/features/Login/ui/MFAForm.tsx
maxto024: Replaced 2 plain buttons with BaseButton

src/pages/MFARecovery/index.tsx
maxto024: Removed unused t variable

src/features/Login/lib/useMFAVerification.ts
maxto024: added the deviceId

src/features/Login/ui/MFAForm.tsx
maxto024: changed

src/features/Login/ui/LoginForm.tsx
maxto024: fixed

src/pages/Login/index.tsx
maxto024: Added documentation comments  explaining why we keep MFAContext (prevents re-renders, password in memory only)

src/features/Login/lib/useMFAVerification.ts
maxto024: I did remove and cleanup

src/features/Login/lib/useMFAVerification.ts
maxto024: fixed push new big commit

sricharan-varanasi (changes requested):
Generally looks good! With the current backend change that includes userID as part of the response, we can update the implementation to derive the key before itself, without passing the password

src/features/Login/lib/useMFAVerification.ts
```diff
@@ -0,0 +1,143 @@
...
+        } = getMfaErrorResult(error, type);
+
+        // Handle session expiry (terminal state)
+        if (expired && !hasSessionExpiredRef.current) {
+          hasSessionExpiredRef.current = true;
+          setIsSessionExpired(true);
+        }
+
+        // Set display error (simple key, or key with attempts if warning threshold)
+        if (remaining !== null && remaining <= WARNING_THRESHOLD && !expired) {
```
sricharan-varanasi: Can we add a Global lockout check, we are currently checking per session limits, but it exactly doesnt notify the user on lockout. we can prioritize global warnings and handle both session limit and lockout warnings.

```
const GLOBAL_WARNING_THRESHOLD = 5; 

if (globalRemaining !== null && globalRemaining <= GLOBAL_WARNING_THRESHOLD && !expired) {
  setDisplayError(`${translationKey}|global|${globalRemaining}`);
} else if (sessionRemaining !== null && sessionRemaining <= SESSION_WARNING_THRESHOLD && !expired) {
  setDisplayError(`${translationKey}|session|${sessionRemaining}`);
} else {
  setDisplayError(translationKey);
}
```

maxto024: Thanks for the comment I pushed a new commit that addresses this 

 The global lockout check is in `MFAVerification.ts` (lines 113-129).

 The logic prioritizes global warnings over session warnings since global lockout is more severe:
```

 if (globalAttemptsRemaining !== null && globalAttemptsRemaining <= WARNING_THRESHOLD && !expired) {
   setDisplayError(`${translationKey}|global|${globalAttemptsRemaining}`);
 } else if (sessionAttemptsRemaining !== null && sessionAttemptsRemaining <= WARNING_THRESHOLD && !expired) {
   setDisplayError(`${translationKey}|session|${sessionAttemptsRemaining}`);
 }
```

 The translation keys globalAttemptsRemaining and sessionAttemptsRemaining are defined in translation.json under the MFA section and handle pluralization automatically via i18next.

src/features/Login/lib/useMFAVerification.ts
divbzero: I’m not seeing `deviceId` in the latest commit (12788c7)?

src/pages/Login/index.tsx
divbzero: I see the comments explaining why we introduced `MFAContext` and don’t know if I’m convinced. I lean towards this TODO described in the comments:

https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/12788c7856b43d21233a7a15ea313f663721f3fc/src/features/Login/lib/MFAContext.tsx#L53

I think you’re right that we need the `userId` in the API response if we want to derive the private key before we `navigate`. This approach would also avoid persisting the password and also avoids introducing a global `MFAProvider` that wraps the entire app. If you agree that this would be a cleaner approach, I think we should consider implementing it now instead of and it as a future `TODO`.

divbzero (commented):
@maxto024 If you agree that it would be better to avoid a global `MFAProvider` that wraps the entire app, I think we should consider implementing this now instead of leaving it as a `TODO` for the future.

src/pages/Login/index.tsx
maxto024: Yep, for sure, since we have the userId now in the response, we will definitely remove the mfacontext and improve this implementation.

src/features/Login/lib/useMFAVerification.ts
maxto024: @divbzero  thanks for catching this but by default it was null but now `Added deviceId: null` to both MFA verification calls. Web clients don't have device IDs (unlike mobile), so we explicitly pass null to match the API contract consistent with our existing logout function.

src/pages/Login/index.tsx
maxto024: @divbzero  Done! Removed MFAContext entirely. Since the API now returns userId and userEmail in the MFA response, we:

  1. Derive private key immediately on Login page (password discarded right after)
  2. Store MFA session for temp jwt in Redux mfa slice (already blacklisted from persistence )
  3. No more global MFAProvider wrapping the app

  This is cleaner  password never leaves the login page

src/features/Login/ui/RecoveryCodeForm.tsx
```diff
@@ -155,38 +158,34 @@ const RecoveryCodeFormComponent = ({
           {/* Show "Back to authenticator" when NOT expired */}
           {!isSessionExpired && (
             <Box display="flex" justifyContent="center">
-              <BaseButton
-                type="button"
-                variant="text"
+              <Text
```
divbzero, line 161: Minor: Was switching from `BaseButton` to `Text` intentional? Seems unusual to have `onClick` on `Text` instead of on a button or link.

src/features/Login/ui/RecoveryCodeForm.tsx
```diff
@@ -155,38 +158,34 @@ const RecoveryCodeFormComponent = ({
...
             </Box>
           )}
 
           {/* Show "Back to Login" ONLY when session expired */}
           {isSessionExpired && (
             <Box display="flex" justifyContent="center">
-              <BaseButton
-                type="button"
-                variant="text"
+              <Text
```
divbzero, line 178: Minor: Same question about `Text` as above.

divbzero (approved):
LGTM. Thank you @maxto024!

(Minor question: Just confirming that replacing `BaseButton` → `Text` was intentional and not a typo.)

sricharan-varanasi (approved):
Looks good to me!

src/features/Login/ui/RecoveryCodeForm.tsx
maxto024: yes it was intentional. baseButton was adding bold, and i didnt want to refactor the baseButton so by changed it to text to match the design.

src/features/Login/ui/RecoveryCodeForm.tsx
maxto024: yep it was intentional baseButton was not matching the font design. I didn't want to refactor that since we are using it in many other places. so converted normal text to match the design.

PR #690, opened 2026-01-12
feat: Fetch flows from server (M2-10229)

divbzero:
### 📝 Description

🔗 [Jira Ticket M2-10229](https://mindlogger.atlassian.net/browse/M2-10229)

Add fetching in-progress flows from server:

- Fetch completed answers from server in `Survey` component
- `includeInProgress=true` when fetching completed answers

Depends on ChildMindInstitute/mindlogger-backend-refactor#1994.

### ✏️ Notes

This PR does not fully implement flow resume yet. The next PR #691 will resolve server vs. local state and check to make sure the activity can be resumed according to schedule.

PR #691, opened 2026-01-16
feat: Resolve server vs. local flow progress (M2-10230)

divbzero:
🔗 [Jira Ticket M2-10230](https://mindlogger.atlassian.net/browse/M2-10230)

There are two main changes in this pull request:

First, update `syncEntity` logic to resolve server vs. local state for flow progress:

1. Pick the state that is farther along in the flow (max `pipelineActivityOrder`).
2. To break ties, prefer more recent state (max `endAt`).

Second, refetch completed entities on survey start unless it is a **Restart**.

Depends on #690, ChildMindInstitute/mindlogger-backend-refactor#1994, and ChildMindInstitute/mindlogger-backend-refactor#2002.


divbzero:
@sricharan-varanasi For easier code review I set this PR to merge into the `M2-10229-flow-resume-fetch-from-server` branch for now. I will switch this PR to merge into `dev` after `M2-10229-flow-resume-fetch-from-server` is merged (#690).

divbzero:
Rebased against `dev` after #690 was merged. No conflicts encountered during rebase.

src/widgets/ActivityGroups/model/hooks/useEntitiesSync.ts
```diff
@@ -45,51 +42,92 @@ export const useEntitiesSync = ({
...
+      // Create or update resumable progress so user can continue where they left off
+      if (isFlow && entity.isFlowCompleted === false) {
+        const flow = applet.activityFlows.find((f) => f.id === entity.id);
+        if (!flow) {
+          console.warn(`[useEntitiesSync] Flow not found for entity ID: ${entity.id}`);
+          return;
+        }
+
+        // activityFlowOrder is 1-indexed, so it equals the 0-indexed position of the next activity
+        const serverPipelineActivityOrder = entity.activityFlowOrder ?? 0;
```
sricharan-varanasi: Was thinking If we dont have activityFlowOrder on the server, we are considering the scenario as the order to be '0'.
can we handle the situation with skipping the sync entirely with local, when server has no data?

src/widgets/ActivityGroups/model/hooks/useEntitiesSync.ts
```diff
@@ -45,51 +42,92 @@ export const useEntitiesSync = ({
...
-            submitId: uuidV4(),
-            context: {
-              summaryData: {},
-            },
+            type: ActivityPipelineType.Flow,
+            currentActivityId: nextActivityId,
+            pipelineActivityOrder,
+            currentActivityStartAt: null,
+            submitId: entity.submitId,
+            startAt: groupProgress?.startAt ?? endAtTimestamp,
```
sricharan-varanasi: Just a quick question : when we sync an in-progress flow to a device that hasn't seen it before, we're setting `startAt` to the last activity's completion time. Is this meant to represent 'flow resumed at X time' rather than 'flow started at X time'? Just want to make sure I understand the intended behavior

sricharan-varanasi (commented):
Can we add test coverage for useEntitiesSync?

src/widgets/ActivityGroups/model/hooks/useEntitiesSync.ts
divbzero: We use this when `isFlow` is `true` and `entity.activityFlowOrder` will typically have a value.

src/widgets/ActivityGroups/model/hooks/useEntitiesSync.ts
divbzero: Thanks @sricharan-varanasi this was a typo. I’ve added tests in commit f51c74b and fixed the code in commit 35d48f1.

src/widgets/ActivityGroups/model/hooks/useEntitiesSync.ts
```diff
@@ -45,51 +40,114 @@ export const useEntitiesSync = ({
...
+        // Skip if local is completed and more recent than server (nothing to update)
+        if ((groupProgress?.endAt ?? 0) > entity.endTime) {
+          return;
+        }
+
+        // Skip if local is in-progress and already ahead of server (nothing to update)
+        if ((groupProgress as FlowProgress)?.pipelineActivityOrder > pipelineActivityOrder) {
+          return;
+        }
+
```
sricharan-varanasi, line 70: Can we check the scenario, if the flow inprogress on web with a submit_id = 'XYZ' is already completed on another device?

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
```diff
@@ -164,7 +165,8 @@ export const ActivityCard = ({ activityListItem }: Props) => {
     return startSurvey({
       activityId: activityListItem.activityId,
       eventId: activityListItem.eventId,
-      targetSubjectId: activityListItem.targetSubject?.id ?? null,
+      targetSubjectId:
+        activityListItem.targetSubject?.id ?? context.applet.respondentMeta?.subjectId ?? null,
       flowId: activityListItem.flowId,
       shouldRestart,
```
sricharan-varanasi, line 169: When user clicks "Resume", this navigates to Survey using local `groupProgress` state without refetching completedEntities from server. can we refetch CompletedEntities before calling Start Survey?

sricharan-varanasi (changes requested):
Looks good chris, can we also add these two checks to avoid stale local data?

src/widgets/ActivityGroups/model/hooks/useEntitiesSync.ts
divbzero: I think the code handles this scenario properly in Case 3 below. I’ve updated the tests as follows:
- Added commit 99e76e2 to always use different server/local submit IDs when testing
- Added commit cda7d2d to test local in-progress flow that was completed on another device

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
divbzero: Thanks for catching this.

Because users can reach the `Survey` not only `ActivityCard` but also other places in the UI, I’m addressing this in commit 3919b07 by adding a redirect within `Survey` if the `currentActivityId` changes after syncing refetching from server.

src/widgets/Survey/ui/index.tsx
```diff
@@ -104,12 +107,38 @@ export const SurveyWidget = (props: Props) => {
...
+
+    const navigateToProps = {
+      appletId,
+      activityId: currentActivityId,
+      entityType: 'flow' as const,
+      eventId,
+      flowId,
+    };
+
+    const screenToNavigate = publicAppletKey
```
sricharan-varanasi: I think the useEffect is triggering whenever `currentActivityId` changes. On final submission in a flow, it becomes undefined and it's redirecting to the first activity, creating an infinite loop . Can we add a guard for this?

sricharan-varanasi (changes requested):

src/widgets/Survey/ui/index.tsx
divbzero: @sricharan-varanasi Thank you. Fixed with commit 4051cfd.

divbzero:
Reverted to using `null` for self-reports in commit 7789cfd and commit 13ad4c0.

sricharan-varanasi (changes requested):
Can we make the Resume button to trigger fresh server fetch? Currently it could resume with stale activity position if mobile completed another activity. Can we invalidate completedEntities query on Resume click or Survey page mount

src/widgets/ActivityGroups/model/services/ActivityGroupsBuildManager.ts
sricharan-varanasi, line 1: The builder is treating ALL Redux groupProgress entries as "in-progress", but entries with endAt set are actually completed flows/activities.
I think we can add `if (groupProgressItem.endAt) continue;` after line 124 to skip completed entries

sricharan-varanasi (changes requested):

divbzero:
@sricharan-varanasi Thank you. Addressed both of your comments with commits d1c5200 and 9e26f31.

Note that instead of invalidating `completedEntities`, I’ve modified the `Survey` to show a loading screen while `completedEntities` is refetching. I think this accomplishes what you had noted.

maxto024 (approved):

divbzero:
Fixed **Restart** of in-progress flow with commits add977b, c9240ea, and bead8bd.

sricharan-varanasi (approved):
Looks good!

PR #697, opened 2026-02-02
feat: `EnableFlowResume` feature flag (M2-10402)

divbzero:
- [x] Tests for the changes have been added

### 📝 Description

🔗 [Jira Ticket M2-10402](https://mindlogger.atlassian.net/browse/M2-10402)

Changes include:

- Add `EnableFlowResume` feature flag
- Use feature flag when fetching completions from server
- Use feature flag when resolving server vs. local state

Depends on #691.

### ✏️ Notes

`EnableFlowResume` can be `true`, `false`, or an array of applet IDs.

PR #699, opened 2026-02-07
Always use `string[]` for `EnableFlowResume` flag

divbzero:

🔗 [Jira Ticket M2-10402](https://mindlogger.atlassian.net/browse/M2-10402)

Changes include:

- `true` → `["*"]` to enable flow resume for all applets
- `false` → `[]` to disable flow resume  for all applets
- `["applet-id-foo", "applet-id-bar"]` to enable flow resume for specific applets 

This complies with LaunchDarkly’s constraint that all JSON variation values must be of the same type.

Related to #697.


maxto024 (approved):

sricharan-varanasi (approved):
Looks good!

PR #700, opened 2026-02-09
fix: Local vs. server sync logic for in-progress flow (M2-10383, M2-10429)

divbzero:
🔗 [Jira Ticket M2-10383](https://mindlogger.atlassian.net/browse/M2-10383)
🔗 [Jira Ticket M2-10429](https://mindlogger.atlassian.net/browse/M2-10429)

Fixed local vs. server activity flow sync logic as follows: 
- Separate logic for same vs. different submissions
- Add missing check if local flow is in-progress and server flow is in-progress
- Add missing check if local flow is in-progress and server flow is completed


sricharan-varanasi (approved):
Looks good!

PR #701, opened 2026-02-11
fix: Redirect when completed on different device (M2-10434, M2-10441)

divbzero:
🔗 [Jira Ticket M2-10434](https://mindlogger.atlassian.net/browse/M2-10434)
🔗 [Jira Ticket M2-10441](https://mindlogger.atlassian.net/browse/M2-10441)

Changes include:

- Modify `useEntitiesSync` to return an array of changed activity/flow IDs
- Redirect `Survey` to activity list if changes indicate a completion on a different device
- Show “**Done!** This survey was already completed on a different device.” on redirect
- Set activity order to display “Activity 1” if user starts activity/flow again

### 📸 Screenshots

Banner displayed if user is redirected back to activity list after completion on a different device:

2026-02-11 Activity Already Completed Banner (M2-10434)

### DIFF
--- a/src/widgets/Survey/ui/index.tsx
+++ b/src/widgets/Survey/ui/index.tsx
@@ -10,6 +10,7 @@ import { FlowProgress, getProgressId } from '~/abstract/lib';
 import { useCompletedEntitiesQuery } from '~/entities/activity';
 import { appletModel } from '~/entities/applet';
 import { useBanners } from '~/entities/banner/model';
+import { PeriodicityType } from '~/entities/event';
 import { AutoCompletionModel } from '~/features/AutoCompletion';
 import {
   SurveyContext,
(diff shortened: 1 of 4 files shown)


src/widgets/Survey/ui/index.tsx
```diff
@@ -142,27 +148,37 @@ export const SurveyWidget = (props: Props) => {
...
+    // Only consider redirect if progress has changed
+    if (!groupProgressChanged) return;
+
+    // If entity was completed on another device, redirect to activity list
+    if (groupProgress?.endAt) {
+      addSuccessBanner(t('additional.activity_already_completed'));
+      navigator.navigate(ROUTES.appletDetails.navigateTo(appletId), { replace: true });
+      return;
+    }
+
```
sricharan-varanasi, line 167: Minor: The previous redirect effect checked flowResumeEnabled before running `if (!flowResumeEnabled || ...)`, but the new effect only checks `groupProgressChanged`. Consider adding `if (!flowResumeEnabled || !groupProgressChanged)` return; for consistency with the feature flag.

sricharan-varanasi (approved):
Looks good chris ! with a minor suggestion to consider. Approved

src/widgets/Survey/ui/index.tsx
divbzero: `!groupProgressChanged` should always be `false` when `!flowResumeEnabled`, but it’s probably better to check explicitly than rely on the implicit behavior. Added `!flowResumeEnabled` check in commit 7725956.

PR #712, opened 2026-03-16
fix: Ux issues - versions (M2-10524, M2-10525, M2-10526, M2-10534, M2-10541, M2-10537, M2-10550)

sricharan-varanasi:
### 📝 Description

🔗 [M2-10524](https://mindlogger.atlassian.net/browse/M2-10524), [M2-10525](https://mindlogger.atlassian.net/browse/M2-10525), [M2-10526](https://mindlogger.atlassian.net/browse/M2-10526), [M2-10534](https://mindlogger.atlassian.net/browse/M2-10534)

Changes include:

- **Latch `storedAppletVersion` with a ref** - The Redux selector for the stored applet version returns `undefined` once `useEntitiesSync` marks an entity as completed (sets `endAt`). This caused a query key change mid-lifecycle, triggering a refetch that blocked the redirect effect and allowed users to re-enter a completed OneTime activity flow. A `useRef` now captures the version on first resolve and keeps it stable for the entire Survey mount.
- **Detect deleted standalone activities on click** - Expanded `fetchFreshFlowData` into a generic `fetchFreshEntityData` that handles both flows and standalone activities. When a standalone activity has been removed by admin, an alert modal is shown and the stale card is refreshed from the API.
- **Defer cache update until alert dismissal** - Previously, the fresh API response was written to the query cache immediately when a deleted flow was detected, causing the card to disappear behind the open modal. The response is now stashed in state and applied only when the user dismisses the alert.
- **Handle deleted flow activities in `ListItemsFactory`** - When a flow still exists but its activities were removed in a newer applet version, `populateActivityFlowFields` no longer throws. It preserves the real `currentActivityId` from progress for in-progress entries (so resume can fetch the activity by its stored version) and renders non-in-progress entries with the flow's own name and

[M2-10524]: https://mindlogger.atlassian.net/browse/M2-10524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[M2-10525]: https://mindlogger.atlassian.net/browse/M2-10525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[M2-10526]: https://mindlogger.atlassian.net/browse/M2-10526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[M2-10534]: https://mindlogger.atlassian.net/browse/M2-10534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
```diff
@@ -153,25 +162,171 @@ export const ActivityCard = ({ activityListItem }: Props) => {
...
-      activityId: activityListItem.activityId,
-      eventId: activityListItem.eventId,
-      targetSubjectId: activityListItem.targetSubject?.id ?? null,
-      flowId: activityListItem.flowId,
-      shouldRestart,
-    });
+    const flowId = activityListItem.flowId;
+
+    if (flowId) {
+      const freshResult = await fetchFreshEntityData(flowId, 'flow').catch(() => undefined);
```
divbzero: I don’t think we should silently ignore errors? Even if we want to continue, could we at least log the error by replacing `.catch(() => undefined)` with `.catch((error) => console.error(error))`?

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
```diff
@@ -153,25 +162,171 @@ export const ActivityCard = ({ activityListItem }: Props) => {
...
+        targetSubjectId: activityListItem.targetSubject?.id ?? null,
+        flowId,
+        shouldRestart: false,
+        freshFlowActivityIds: freshResult.activityIds,
+      });
+    } else {
+      // For standalone activities, check if the activity still exists
+      const freshResult = await fetchFreshEntityData(activityListItem.activityId, 'activity').catch(
+        () => undefined,
+      );
```
divbzero, line 193: Same as above.

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
```diff
@@ -153,25 +162,171 @@ export const ActivityCard = ({ activityListItem }: Props) => {
...
+
+  const restartActivity = useCallback(async () => {
+    const flowId = activityListItem.flowId;
+
+    if (!isEntitySupported) {
+      return openStoreLink();
+    }
+
+    if (flowId) {
+      const freshResult = await fetchFreshEntityData(flowId, 'flow').catch(() => undefined);
```
divbzero: Same as above.

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
```diff
@@ -153,25 +162,171 @@ export const ActivityCard = ({ activityListItem }: Props) => {
...
+        targetSubjectId: activityListItem.targetSubject?.id ?? null,
+        flowId,
+        shouldRestart: true,
+        freshFlowActivityIds: freshResult.activityIds,
+      });
+    } else {
+      // For standalone activities, check if the activity still exists
+      const freshResult = await fetchFreshEntityData(activityListItem.activityId, 'activity').catch(
+        () => undefined,
+      );
```
divbzero, line 337: Same as above.

divbzero (approved):
Thanks @sricharan-varanasi for all the fixes here! Just noticed a few `.catch(() => undefined)` where we should probably log the errors before continuing.

src/widgets/ActivityGroups/ui/ActivityCard/index.tsx
sricharan-varanasi: Resolved, @divbzero , thanks for noticing

PR #716, opened 2026-04-02
fix: include start/end time when validating availability window

adeiji:
### 📝 Description
🔗 [Jira Ticket M2-8773](https://mindlogger.atlassian.net/browse/M2-8773)

Fixed a bug where scheduled events were not showing on the end date. For example, if the scheduled event's end date was April 1, 2026 @ 11:59 PM and on April 1, 2026 @ 5 PM the user is viewing available events, the event would not show up.  Now, when viewing the event on the last day:

- The event shows on the dashboard

### 🪤 Peer Testing

- Create a scheduled event in the admin app with an end date of today, and the end time some time in the future
- Open the web app, and select the applet with the event

**Expected Outcome:** The event should be visible

- Go back to the admin app, and set the end time to an earlier time.
- Go back to the web app and refresh.

**Expected outcome**: The event should not be visible

### ✏️ Notes

Validation was comparing only the date component and ignoring the time component.

### DIFF
--- a/src/abstract/lib/GroupBuilder/GroupUtility.ts
+++ b/src/abstract/lib/GroupBuilder/GroupUtility.ts
@@ -264,9 +264,6 @@ export class GroupUtility {
           )
         : (endDate ?? undefined);
 
-    console.log('from', from);
-    console.log('to', to);
-
     return this.isInInterval(
       {
         from,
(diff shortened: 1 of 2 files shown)


src/abstract/lib/GroupBuilder/GroupUtility.ts
```diff
@@ -238,13 +238,36 @@ export class GroupUtility {
 
   public isInsideValidDatesInterval(event: ScheduleEvent) {
     const { startDate, endDate } = event.availability;
+    const { timeFrom, timeTo } = event.availability;
```
sricharan-varanasi: nit: can we do a single destructure instead of two?

sricharan-varanasi (changes requested):
Looks good, Ade. 
Can we add tests for this case?

sricharan-varanasi (approved):

PR #751, opened 2026-06-18
chore: Upgrade Node `20.11.*` → `24.*.*`

divbzero:
### 📝 Description

We already specify Node 24 in our other repositories. Upgrading here for consistency.

Changes include:
- Upgrade Node `20.11.*` → `24.*.*` for NVM
- Upgrade Node `20.11.0` → `24.*.*` for Docker
- Clean up README and GitHub workflows

### ✏️ Notes

I’ve been using Node 24 for awhile locally and haven’t encountered any issues.


sricharan-varanasi (approved):
Looks good to me, Chris. One small thought: node:24 in the Dockerfile floats on the major, do we need to pin it to a specific version?

PR #763, opened 2026-08-14
feat: Keep the session alive on Web (M2-11003)

sricharan-varanasi:
### 📝 Description

🔗 [Jira Ticket M2-11003](https://mindlogger.atlassian.net/browse/M2-11003)

The inactivity logout runs on a `setTimeout` each tab owns on its own, so it fires late, early, or not at all. Two tabs count separately and whichever fires first logs you out of both. Tokens live in local storage, so closing the browser ends nothing - you come back looking signed in until the first request fails.

This replaces the countdown with a wall clock every tab shares.

Changes include:

- `lastActivityAt` in plain local storage. Every tab writes it; the timer re-reads it when it fires instead of logging out blind, so activity in one tab keeps them all alive.
- A re-check on `visibilitychange`, so a tab the browser suspended finds out on focus rather than whenever its timer happens to run.
- A boot check that ends a session already past its deadline, clearing exactly what a logout clears.
- Single-flight token refresh. Three requests that 401 together currently trigger three refreshes; now one.
- Proactive refresh armed off the access token's `exp`. Behind `enableSessionKeepAlive`, off by default.
- Idle timeout moved to 30 minutes and made configurable via `VITE_IDLE_TIMEOUT_MIN`.
- Tests for the 401 interceptor, which had none.

### 🪤 Peer Testing

Needs short timeouts in `.env` and a dev server restart:

```
VITE_IDLE_TIMEOUT_MIN=3
VITE_REFRESH_LEAD_SEC=20
```

Use `chrome://discards` → Freeze for the frozen-tab case. Backgrounding a tab is not the same thing.

- **Two tabs signed in. Work continuously in tab A past the timeout, leave tab B untouched in the background**

    **Expected outcome:** neither logs out.

- **Leave both tabs idle past the timeout**

    **Expected outcome:** both land on the login page

- **Freeze tab B, keep working in tab A past the timeout, then focus tab B**

    **Expected outcome:** tab B stays signed in and does not log tab A out

- **Sign in, close the browser, come back after the timeout**

    **Expected outcome:** the login page, not a signed-in page that fails on the first click

- **Start an activity, answer a few items, then idle out**

    **Expected outcome:** unchanged from today - logged out, answers cleared

- **Sign in with MFA and type the code wrong**

    **Expected outcome:** "incorrect code" on the MFA screen, not a bounce back to email and password

### ✏️ Notes

- **The idle timer ships unflagged.** 
- **`InactivityTracker.tsx` is deleted, `useIdleTimer` is not.** `PassingScreen` still uses it for the applet-configured auto-completion timer.

### Functionality

- [x] The feature behaves correctly in practice and fulfills the intended business purpose
- [x] The implementation accounts for edge cases, avoids subtle logical errors, and handles somewhat rare failure states (e.g. offline mode for mobile, 3rd party being down, etc)

### Testing

- [x] Verify there are automated tests added that meaningfully cover critical behavior and failure cases
- [x] Code coverage does not go down as result of this change
- [x] Test suite passes

### Security & Data Privacy

- [x] Verify there is no chance we would accidentally log PII to application logs
- [x] Verify this addition does not materially affect our security attack surface, and if so it has undergone security review
- [x] All inputs are sanitized
- [x] New dependencies are well maintained, have significant justification for being added to the project, and are documented in the Curious [open source credit page](https://mindlogger.atlassian.net/jira/servicedesk/projects/MLA/knowledge/articles/340623543?spaceKey=MLA)

### Logging/Monitoring

- [ ] Logging is implemented for this change such that you could troubleshoot this feature in production
- [ ] The change/feature is able to be monitored in production

### Performance

- [x] This change does not introduce n+1 queries or other performance issues within our expected scale (e.g. missing indexes on frequently queried columns, frequently updating tables that are accessed often)

### Readability

- [x] All commented out code is removed
- [x] Debugging code including extraneous log lines are removed
- [x] Code is easy to understand through naming and structure; comments explain intent or non‑obvious decisions

### Change Safety

- [x] ~~Backend changes are backwards compatible with old clients, or it is well known they are not and a deployment/rollout plan is in place. This include backend changes being compatible with old mobile app versions, as well as applet versioning within Curious.~~
- [x] ~~Destructive database migrations are rolled out in stages. For example, renaming a column means adding a new column and migrating the existing data to that columns in one deployment. Then monitoring to ensure that field isn't used, and finally removing that old column in a separate deployment.~~


divbzero:
Superseded by #764.

PR #764, opened 2026-08-18
feat: Session tab sync and keep session alive (M2-11004,M2-11005)

sricharan-varanasi:
### 📝 Description

🔗 [Jira Ticket M2-11004](https://mindlogger.atlassian.net/browse/M2-11004)
🔗 [Jira Ticket M2-11005](https://mindlogger.atlassian.net/browse/M2-11005)

Today each tab runs its own session in isolation. Three problems follow: once the backend
enables refresh-token rotation, a second tab spending a superseded token revokes the whole
family and logs everyone out; a logout in one tab leaves the others holding a dead token
until their next 401; and a tab on the login page never notices a sign-in elsewhere.

This also adds the idle warning. Today the session simply ends at 30 minutes with no notice,
which can drop a respondent mid-activity with answers part-filled and nothing explaining why.

Changes include:

- A `BroadcastChannel` with four messages - `SESSION_REQUEST`, `SESSION_STATE`, `TOKENS_UPDATED`, `LOGOUT`
- Tabs matched by the refresh token's `family` claim (`jti` fallback), so another account's traffic is ignored
- A signed-in tab takes a sibling's rotated tokens in place and re-arms its timer off the new expiry
- A tab waking from sleep asks for current tokens before spending its own
- Logout fans out; the receiving tab skips the API call and the duplicate Mixpanel event
- A signed-out tab reloads to join a session, and reloads itself when no tab is left to answer

**M2-11005**

- A warning modal at 25 minutes idle, counting down the last 5 before logout
- **Stay logged in** keeps the session; **Log out** ends it. Closing the modal any way keeps it
- Activity pauses while the modal is open, so moving the mouse toward the buttons cannot answer it
- Countdown length set by `VITE_IDLE_WARNING_MIN`, capped at half the idle timeout
- Answering in one tab closes the modal in the others, off the shared activity clock
- A signed-in tab visiting `/login` is sent into the app instead of being shown the login form

### 🪤 Peer Testing

Needs `enableSessionKeepAlive` **on** in LaunchDarkly. No new dependencies.

- **Sign in on Tab A with a second tab already open on the login page.**

    **Expected outcome:** the second tab lands in the session on its own, without being touched.

- **With two tabs signed in, wait past the refresh lead.**

    **Expected outcome:** one `POST /auth/token/refresh` per cycle, not two, and both tabs keep working.

- **Log out in Tab A.**

    **Expected outcome:** Tab B lands on the login page immediately, and only one `POST /auth/logout` is sent.

- **Freeze Tab B, log out in Tab A, then unfreeze B.**

    **Expected outcome:** B tears down when it comes back into focus.

- **Sign in, open a second tab, close the first, then focus the second on the login page.**

    **Expected outcome:** it reloads into the session, exactly once.

- **Sign in on Tab A, let Tab B join, then complete and submit an activity in Tab B.**

    **Expected outcome:** the answer submits. 

**M2-11005**
  
The warning needs a build with shortened timings - on this branch it only appears after 25 real
minutes of untouched idling.

- **Sign in, then leave the machine completely alone until the modal appears.**

    **Expected outcome:** it counts down, and the count matches the time left.

- **Move the mouse without clicking anything.**

    **Expected outcome:** the countdown keeps running. Activity does not dismiss it.

- **Click Stay logged in.**

    **Expected outcome:** the modal closes and the session survives past its original deadline.

- **Trigger it again and let it reach 0:00.**

    **Expected outcome:** logged out, landing on the login page.

- **Open two tabs, let both show the modal, click Stay logged in in one.**

    **Expected outcome:** the other tab's modal closes within a second.

- **Sign in, then type `/login` in the address bar.**

    **Expected outcome:** you land in the app, not on a login form.

### ✏️ Notes
- A signed-out tab **reloads** rather than writing tokens in place. `userPrivateKey` is derived from the password and lives in the same import-time snapshot as the tokens, so writing them in would make the tab look signed in and then fail at answer-submit. This is the one place web diverges from admin.
- Tokens travel over the channel in plaintext. It's same-origin only, so the same exposure as the localStorage they already sit in.
- 855 tests pass, lint clean, build clean.

### Functionality

- [x] The feature behaves correctly in practice and fulfills the intended business purpose
- [x] The implementation accounts for edge cases, avoids subtle logical errors, and handles somewhat rare failure states (e.g. offline mode for mobile, 3rd party being down, etc)

### Testing

- [x] Verify there are automated tests added that meaningfully cover critical behavior and failure cases
- [x] Code coverage does not go down as result of this change
- [x] Test suite passes

### Security & Data Privacy

- [x] Verify there is no chance we would accidentally log PII to application logs
- [x] Verify this addition does not materially affect our security attack surface, and if so it has undergone security review
- [x] All inputs are sanitized
- [x] New dependencies are well maintained, have significant justification for being added to the project, and are documented in the Curious [open source credit page](https://mindlogger.atlassian.net/jira/servicedesk/projects/MLA/knowledge/articles/340623543?spaceKey=MLA)

### Logging/Monitoring

- [ ] Logging is implemented for this change such that you could troubleshoot this feature in production
- [ ] The change/feature is able to be monitored in production

### Performance

- [x] This change does not introduce n+1 queries or other performance issues within our expected scale (e.g. missing indexes on frequently queried columns, frequently updating tables that are accessed often)

### Readability

- [x] All commented out code is removed
- [x] Debugging code including extraneous log lines are removed
- [x] Code is easy to understand through naming and structure; comments explain intent or non‑obvious decisions

### Change Safety

- [x] ~~Backend changes are backwards compatible with old clients, or it is well known they are not and a deployment/rollout plan is in place. This include backend changes being compatible with old mobile app versions, as well as applet versioning within Curious.~~
- [x] ~~Destructive database migrations are rolled out in stages. For example, renaming a column means adding a new column and migrating the existing data to that columns in one deployment. Then monitoring to ensure that field isn't used, and finally removing that old column in a separate deployment.~~


src/app/model/clearStaleSession.ts
```diff
@@ -0,0 +1,48 @@
...
+// over on its first request.
+export const clearStaleSession = () => {
+  const lastActivityAt = getLastActivityAt();
+  // Nothing has tracked this session, so there is no deadline to judge it against. Leave it to the
+  // usual 401 path.
+  if (!lastActivityAt) return;
+
+  if (Date.now() - lastActivityAt < resolveSessionConfig().idleTimeoutMs) return;
+
+  // Everything a logout clears, so a session that ends here ends the same way as one ended by hand.
```
divbzero: [**Important**] Also invalidate tokens on server like we do in `useLogout.ts`? Similar to the comment on `clearStaleSession.ts` in ChildMindInstitute/mindlogger-admin#2251.

src/shared/api/services/axios.ts
```diff
@@ -32,11 +39,52 @@ axiosService.interceptors.request.use(
...
+  const tokens = secureTokensStorage.getTokens();
+
+  if (!tokens?.refreshToken) {
+    throw new Error('No refresh token to refresh with.');
+  }
+
+  const { data } = await authorizationService.refreshToken({
+    refreshToken: tokens.refreshToken,
+  });
+
```
divbzero: [**Important**] Check if user has logged out while awaiting the refreshed token? Similar to the comment on `api.utils.ts` in https://github.com/ChildMindInstitute/mindlogger-admin/pull/2251.

divbzero (commented):
@sricharan-varanasi Looks good overall! Please see comments below, similar to comments in ChildMindInstitute/mindlogger-admin#2251.

src/shared/utils/session/leaveEndedSession.ts
```diff
@@ -0,0 +1,18 @@
...
+// screen is the old user's. So it drops what is its own and reloads, leaving a note for the boot on
+// the way back in.
+export const leaveEndedSession = () => {
+  // Per tab, so clearing it takes nothing from the live session. It survives a reload, which is why
+  // the state of a session that has ended has to be dropped here rather than left to the reload.
+  sessionStorage.clear();
+
+  // Set after the clear, and read on the way back in. Without it the reload would read the tokens
+  // the browser now holds and walk straight into a session this tab was never signed in to.
+  sessionStorage.setItem(SESSION_ENDED_KEY, 'true');
```
divbzero: This part makes sense, but I think we need a corresponding place where we unset it after a user logs in again.

We could probably add this in the `onLoginSuccess` handler:

```ts
sessionStorage.removeItem(SESSION_ENDED_KEY);
```

src/widgets/SessionKeepAlive/lib/useSessionAdoption.ts
```diff
@@ -0,0 +1,113 @@
...
+
+      raiseBanner();
+    };
+
+    const armFallback = () => {
+      clearTimeout(fallbackTimer);
+      fallbackTimer = setTimeout(raiseBannerFromClock, SESSION_REQUEST_WINDOW_MS);
+    };
+
+    const unsubscribe = subscribeSessionSync((message) => {
```
divbzero: What happens if we’ve called `raiseBanner()` and set `SESSION_ELSEWHERE_KEY` but the user logs out in another tab? I think we need to listen for the `LOGOUT` event and unset `SESSION_ELSEWHERE_KEY` here:

```ts
if (message.type == 'LOGOUT') {
  hasRaised = false;
  sessionStorage.removeItem(SESSION_ELSEWHERE_KEY);
}
```

divbzero: Looks like @sricharan-varanasi has addressed this with `clearBanner()` in commit 662d19f.

divbzero:
Thank you @sricharan-varanasi. No new concerns with the last two commits.

src/shared/api/services/axios.ts
sricharan-varanasi: thank you chris, addressed in 4128c381

src/app/model/clearStaleSession.ts
sricharan-varanasi: thank you chris, addressed in 69ccb80e

src/shared/utils/session/leaveEndedSession.ts
sricharan-varanasi: thank you chris, addressed in c13e7279

src/shared/api/services/axios.ts
divbzero: Maybe move up to here line 65:

```ts
const sessionId = getSessionId();
```

src/shared/api/services/axios.ts
```diff
@@ -54,6 +54,11 @@ const requestNewTokens = async () => {
     refreshToken: tokens.refreshToken,
   });
 
+  // A logout landed while this was in flight, so storing would put the session back on its feet.
+  if (!secureTokensStorage.getTokens()?.refreshToken) {
```
divbzero: And change this condition to the following?

```ts
!secureTokensStorage.getTokens()?.refreshToken || getSessionId !== sessionId
```

src/app/model/clearStaleSession.ts
```diff
@@ -32,14 +34,24 @@ const clearPersistedSlices = () => {
...
   if (Date.now() - lastActivityAt < resolveSessionConfig().idleTimeoutMs) return;
 
+  const tokens = secureTokensStorage.getTokens();
+  // Today the refresh token expires as the session goes stale, so this is skipped. It matters if
+  // the idle timeout is ever shortened below the token's lifetime, leaving a live one behind.
+  const refreshTokenExpiresAt = getTokenExpiration(tokens?.refreshToken);
+
+  if (tokens?.accessToken && refreshTokenExpiresAt !== null && refreshTokenExpiresAt > Date.now()) {
+    // Revokes the whole family, so the refresh token goes with it.
+    await authorizationService.logout({ accessToken: tokens.accessToken }).catch(() => undefined);
```
divbzero: It might be better if we switched from `authorizationService.logout` which calls `POST /auth/logout` to a new `authorizationService.logout2` method that hits `POST /auth/logout2`? I think `POST /auth/logout2` accepts `refreshToken` which has a longer expiration than `POST /auth/logout` which accepts `accessToken`.

divbzero: Also, would it make sense to remove the `await`? Just send out the request to revoke but proceed with mounting the app without waiting?

divbzero (approved):
@sricharan-varanasi A few more questions about the most recent commits, but otherwise looks ready to merge. Thank you!

src/shared/api/services/axios.ts
sricharan-varanasi: Thank you @divbzero . Session id is now read before the request and compared
after. Used getSessionId() rather than getSessionId in the comparison.

src/app/model/clearStaleSession.ts
sricharan-varanasi: Thank you @divbzero . We were checking the refresh token's expiry but sending the access token, which is always dead by the time a session goes stale, so the revoke just 401'd and nothing was ever revoked. Switched to logout2 in c85e7435.

the request interceptor attaches the access token to every outgoing request, so it was overwriting the refresh token we'd set. It now leaves an Authorization header alone if the caller set one (8dd3d4fe). Also marked the revoke with retry: true so a refusal falls straight through , otherwise a failed revoke would try to refresh the dead session and fire a global logout during boot.

On the await: removed in d40a5932. Safe now because the token travels on the request rather than being read from the store as it goes out, so the clear can't overtake it.

src/app/model/clearStaleSession.ts
divbzero: After removing the `await` below, `clearStaleSession` doesn’t have to be `async` anymore?

sricharan-varanasi: thank you chris, addressed in f8ae7c59

src/entities/banner/model/slice.ts
divbzero: I noticed we aren’t filtering by `key` which could result in duplicate banners in some cases. Maybe before `.push(payload)` we `.filter(...)` using the same logic we have below in  `removeBanner`:

```ts
state.banners = state.banners.filter(({ key }) => key !== payload.key);
state.banners.push(payload);
```

divbzero (approved):
@sricharan-varanasi One small suggestion. The rest looks good!

src/entities/banner/model/slice.ts
sricharan-varanasi: addressed in 2b281ad, thank you

